PR #4, opened 2017-04-25
SaveDir configureable

adamschaub:
Fix the thing I broke.

Create the testing directory relative to where the test is being run. Better ensures permissions will not be a blocking factor.


adamschaub:
One limitation with this. If running sigma in a container, the saveDir for docker tasker will have to match the mounted directory from the host machine exactly.

PR #7, opened 2017-04-27
Godoc

stephenrlouie:
### Change log

- Adds comments on exported functions, variables, and package names.
- Formatting on import statements to follow developer.md (`built-in, external, internal`)


api/v1/server/server.go
```diff
@@ -29,13 +32,14 @@ type Server struct {
 	quit chan struct{}
 }
 
+//New creates a new http server instnace by taking in a url to listen on and a database object
```
meganokeefe: instnace -> instance

task/docker.go
```diff
@@ -32,16 +32,19 @@ var (
 	OutputDirDefault = "/var/lib/sigma"
 )
 
+//DockerConfig defines the docker tasker config
 type DockerConfig struct {
 	SaveDir string
 }
 
+//DockerTasker
```
meganokeefe: need a comment here that isn't just the name of the struct

task/task.go
```diff
@@ -1,8 +1,10 @@
+//Package task deifnes the task interface that is required to run and monitor a "task".
```
meganokeefe: deifnes -> defines

PR #14, opened 2017-04-28
main at the top level

stephenrlouie:
### Change log

Bringing the `main.go` file to the root dir. Separated the cmd definition into the `cmd` package

### Notes

For easier building, go-thic for `go get`


adamschaub:
So tidy, I love it.

PR #15, opened 2017-04-28
adding dev guide

stephenrlouie:
### Change log

moving the developer guide to pretty docs


_docs/developer-guide.md
```diff
@@ -0,0 +1,118 @@
+# Developer Guide
```
meganokeefe: The doc has to have a jekyll header so that it knows what the URL should be, add it to the Docs page, etc. 

ie. 
```
---
layout: page
title: Developer Guide
permalink: /docs/developer-guide/
---
```

See the other `_docs` for examples.

meganokeefe: You can also remove the main header `#Developer Guide` because Jekyll will take your `page.Title` and put it at the top in big font.

_docs/developer-guide.md
```diff
@@ -0,0 +1,121 @@
...
+## Error Handling
+- Pass errors as far up until a container boundary is met.
+- If not recoverable log as close to the container boundary as possible, providing stack trace / relevant context for the error and exit cleanly. (caller, transaction, etc)
+- If recoverable, warning log and move on.
+- Use idiomatic Go error handling, use types where possible. Never `_` when an `err` is a potential return. Always immediately check `if err != nil`
+- NEVER `os.Exit()`, always return up to `cmd/main` and let it exit.
+
+
+## Logging
+- Appropriately level your logs (Debug, Info, Warning, Error, Fatal, Panic)
```
adamschaub: In the future, it would be good to specify what should fall under each log level. We aren't picky about it right now though

_docs/developer-guide.md
```diff
@@ -0,0 +1,121 @@
...
+- Ensure logs have enough context to be useful
```
adamschaub: Define what enough to be useful is

_docs/developer-guide.md
```diff
@@ -0,0 +1,121 @@
...
+- Logs must be utilized for debugging in a production environment
```
adamschaub: I'm not sure what this means

_docs/developer-guide.md
```diff
@@ -0,0 +1,121 @@
...
+- Use `WithFields` or `WithField` to add any additional variables to the log. Never use the string format operators to add variables into the logs and capitalize all Field Keys. IE: `log.WithFields(log.Fields{"Test": "Value", "Test2": :foobar,}).Info("My logging message")`
```
adamschaub: I don't agree with the no string operators sentiment. Adding fields is nice, but a filled-in log line can be easier for a person to understand.

_docs/developer-guide.md
```diff
@@ -0,0 +1,121 @@
...
+- Imports are separated into three groups, `built-in packages`, new line, `external packages`, new line, `internal packages`
+- Simple descriptive variable names. camelCase if you have to.
+
+
+## Adding Dependencies
+- Sigma leverages [GoVendor](https://github.com/kardianos/govendor). A package can be added to the project by running `govendor fetch <package_name>`. Ensure that the `vendor/vendor.json` file contains your package and the `vendor/` directory has the project.
+
+
+## Error Handling
+- Pass errors as far up until a container boundary is met.
```
adamschaub: Meaning every error should get elevated as a service level error? I was under the impression that that was only for service-ending errors.

adamschaub: Container as in like a struct/interface? Not sure what this is implying.

_docs/developer-guide.md
stephenrlouie: want to flush this out RN?

_docs/developer-guide.md
stephenrlouie: probably combined with *enough to be useful*

_docs/developer-guide.md
stephenrlouie: as close to the top of the call stack.

PR #24, opened 2017-04-28
Return ID list instead of struct

stephenrlouie:
### Change log

Implements query functions to return `[]string` which is the IDs instead of the entire struct. performance optimization

### Notes

I have not implemented any tests here, plan on following up for an entire CB test PR, that will test the entire interface against a CB container. (need to learn how to replicate jenkins side car container on travis)


database/couchbase.go
```diff
@@ -517,3 +547,13 @@ func replaceErr(err error) error {
 		return err
 	}
 }
+
+func parseIds(rows gocb.QueryResults) []string {
+	type retId map[string]string
```
meganokeefe: reason why you aren't just saying `var retS map[string]string`?

meganokeefe:
Very nice, I like this change. LGTM.

PR #27, opened 2017-05-01
Filtering out . 

stephenrlouie:
### Change log

- Filters out `.` from Plugin & Service `Names`
- Filters out `.` from any Input & Output `Names`

Ensures splits in [manager](https://github.com/sigma-dev/sigma/blob/master/manager/manager.go#L400) are not thrown off by user input


api/v1/server/server.go
```diff
@@ -357,3 +358,40 @@ func (s *Server) UpdateService(w http.ResponseWriter, r *http.Request) {
...
+			if strings.Contains(i.Name, ".") {
+				return true
+			}
+		}
+		for _, o := range s.Output {
+			if strings.Contains(o.Name, ".") {
+				return true
+			}
+		}
+	case apiclient.Plugin:
```
meganokeefe: add a default error case to the switch?

api/v1/server/server_test.go
```diff
@@ -292,6 +306,84 @@ var fullService = apiclient.Service{
...
+		apiclient.ServiceIo{
+			Name:  "in.1",
+			Type_: "in1",
+			Value: nil,
+			From:  nil,
+		},
+	},
+}
+
+var periodNamePlugin = apiclient.Plugin{
```
meganokeefe: lolz

meganokeefe:
Question, why are we kicking out plugins with periods in them? As opposed to doing a find and replace on the `Names` and just inserting the plugin anyway. 

But I see the merit of throwing an error, for clarity.

stephenrlouie:
manager implies meaning based off a `split(str, ".")` [code](https://github.com/sigma-dev/sigma/blob/master/manager/manager.go#L407-L412)

api/v1/server/server.go
stephenrlouie: yea, ill just return false, which will trigger the `StatusBadRequest`

PR #29, opened 2017-05-01
adding cron and functional test placeholder

stephenrlouie:
### Change log

- Adds a placeholder script for functional tests
- Will set up travis to run the cron job once this merges. Will have a container for each cron build


.travis.yml
```diff
@@ -8,6 +8,7 @@ env:
   global:
     - PROD_REPO=sigmadev/sigma
```
meganokeefe: what do these repos refer to? Is `sigmadev` the org? If so, why isn't it `sigma-dev?` 

and is `sigma` the repo? in which case, will `sigma-dev` and `sigma-cron` be repos in the org, eventually?

meganokeefe: wait do these refer to docker

stephenrlouie: yup: [DockerHub](https://hub.docker.com/u/sigmadev/)

PR #30, opened 2017-05-01
Sites

meganokeefe:
# Sigma Sites

## Changelog 
- `model.go` definition for Site. every Service gets a Site name.
- Database functions for Sites
- API endpoints and functions for Sites: insert site, get site, get all sites, update site, delete site. the API server `AddSite()` function does a `validateSite()` to check the site's children.
- Manager helper `choosePhysicalSitel()` is used when someone wants to run a Service on a Logical site. It gathers all the Physical child sites of that Logical sites and chooses one at random to pass to Tasker. Physical sites are passed to Tasker as-is.
- Manager automatically adds site `local` to the DB, on startup. 

- updated API spec for the docs site 
- added documentation for Sites to the docs site 

## Notes 

My initial notes/ user stories, some of this stuff has since changed:
https://github.com/sigma-dev/sigma/blob/sites/docs/sites-notes.md

## Todo 

Sites-related stuff not included in this PR: 

1. DockerTasker configuration to explicitly run locally given a `local` site for a service
2. DockerTasker configuration/authentication to allow remote clients. aka non-`local` Sites for DockerTasker
3. Future tasker implementations that have an idea of Sites -- ie. K8s, Nomad -- should ingest the `ServiceOperation.Site` and use it. They should also know how to process Site==`local`.

Also, will need to update the docs again after this lands, since some stuff has changed since I added the Sites 101 doc.


docs/sites-notes.md
```diff
@@ -0,0 +1,70 @@
...
+              SFO (P)
+              SJC (P)
+            SOCAL (L)
+              LA (P)
+              SANDIEGO (P)
+
+
+- Sites can be physical (P) or logical (L). Logical sites like "US-EAST" encapsulate child sites, and are not runnable.
+Ie. you can't directly deploy a Service onto US-EAST.
+- The lowest level of the sites tree MUST be physical. Physical sites are runnable sites-- for instance, Nomad NodeIDs, K8s cluster IDs, etc. "Physical" is anything that comes upstream through Mu Ctrl and has a site made out of it.
```
stephenrlouie: no mention of mu. Users must make physical sites that are runnables from the tasker. We don't support a way to fetch.

stephenrlouie: However you configure your task runners, you need to feed that into the sitesAPI as physical sites.

docs/sites-notes.md
```diff
@@ -0,0 +1,70 @@
...
+- Let's say the Operator is using a NomadTasker and leveraging the Sigma UI, which is connected to the API.
+- Operator can see in the UI that they have three Physical (P) sites, each a Nomad Node ID that has arrived upstream from the nomad tasker runner.
+The sites have uuids 1001, 1002, and 1003.
+- Operator knows each of these sites are in MASS. and wants to group them into a site.
+- Operator clicks on the "Create a Site" button and checks "1001" "1002" and "1003" as Children to add. They give the site a name - MASS - and Sigma takes care of the rest.
+- Now back in the UI graphic, MASS appears as a parent of those three sites.   
+
+**Running a Service on a Site**
+- Continue where we left off in the previous story. MASS is a Logical site over three Physical sites.
+- The Operator wants to run a Service on MASS. The Service is a Video Transcoder container that has heavy compute needs.
```
stephenrlouie: maybe another random example.

docs/sites-notes.md
```diff
@@ -0,0 +1,70 @@
...
+      Site struct {
+        Id string           //autogen uuid, used for internal sigma use
+        Name string         //name for this site. either user defines it or it's the upstream ID. what users will use in Service defns. must be unique.
+        Physical bool       // True==this site is runnable (came upstream and refers to an actual cluster or node). False== this site came from API, is a logical grouping.
+        Children []string   //list of Site.Name - can be both physical and logical. DB will check to validate these children (ensure they exist already, ensure all children are parentless)
+      }
+
+## Changes to Sigma
+
+(remember a P site name can refer to: NomadNodeID, K8s Cluster Name)
```
stephenrlouie: it is the native lowest level id for that tasker. aka: a TaskRunnerID whatever that means for the implementation of tasker

docs/sites-notes.md
```diff
@@ -0,0 +1,81 @@
...
+**Notes:**
+- the leaves (lowest level) of a Site tree must be Physical.
+- a Site can have up to one parent. A Site is not allowed to have multiple parents.
+
+
+## Site User Stories / Flow
+
+#### 1. Adding a Logical Site to Sigma
+
+- Let's say the Operator is using a NomadTasker and leveraging the Sigma UI, which is connected to the API.
```
stephenrlouie: don't need 
> which is connected to the API

docs/sites-notes.md
```diff
@@ -0,0 +1,81 @@
...
+- Operator can see in the UI that they have three Physical (P) sites, each a Nomad Node ID that has arrived upstream from the nomad tasker runner.
```
stephenrlouie: task runners are not responsible for putting physical sites there, they must be feed into Sigma via the sites API

docs/sites-notes.md
```diff
@@ -0,0 +1,81 @@
...
+- Operator clicks on the "Create a Site" button and checks "1001" "1002" and "1003" as Children to add. They give the site a name - MASS - and Sigma takes care of the rest.
+- Now back in the UI graphic, MASS appears as a parent of those three sites. A Logical site is born!  
+
+#### 2. Running a Service on a Logical Site
+
+- Continue where we left off with Story #1. MASS is a Logical site, containing three Physical sites.
+- The Operator wants to run a Service on MAS- some container with a heavy compute process.
+- The Operator creates a Service definition and supplies the MASS site name.
+- Internally, Sigma fetches the MASS Logical site and all its Physical children. Sigma will choose a random Physical Site . It chooses Site 1002.
+- The Operator can now see in the UI that their Transcoder Service is running on Site 1002- (ie. Nomad has deployed the container onto a Node with id=1002.)
```
stephenrlouie: transcoder?

docs/sites-notes.md
```diff
@@ -0,0 +1,81 @@
...
+
+#### 3. Updating a Service to Run on a New Site
+
+- Continue where we left off with Story #2. our Service is running on Site 1002.
+- The Operator gets word that snow and high winds are about to hit the town where Site 1002 is located. Worried about a power outage, the Operator decides to manually transfer their Service from 1002 to 1003.
+- They update their Service definition accordingly in the UI. Instead of assigning the Service to the MASS (Logical) site, they specify the 1003 (Physical) site instead.
```
stephenrlouie: I don't think user interaction will be required. Sigma should be smart enough to see when a service doesn't work on the requested physical site, it will pick another. So they could just assign east, if the blizzard takes out 1002, sigma should know this and move it. (assuming long running job and all)

docs/sites-notes.md
```diff
@@ -0,0 +1,92 @@
...
+- Let's say the Operator is using a NomadTasker and leveraging the Sigma UI.
+- Operator can see in the UI that they already have three Physical (P) sites, which they added earlier. The sites have uuids 1001, 1002, and 1003.
+- Operator knows each of these sites are in MASS. and wants to group them into a site.
+- Operator clicks on the "Create a Site" button and checks "1001" "1002" and "1003" as Children to add. They give the site a name - MASS - and Sigma takes care of the rest.
+- Now back in the UI graphic, MASS appears as a parent of those three sites. A Logical site is born!  
+
+#### 2. Running a Service on a Logical Site
+
+- Continue where we left off with Story #1. MASS is a Logical site, containing three Physical sites.
+- The Operator wants to run a Service on MAS- some container with some long-running, heavy-compute process.
```
stephenrlouie: missing an `S`

PR #32, opened 2017-05-01
Configuration

adamschaub:
### PR checklist

- [x] Provide default configurations for each configureable component
- [x] Add flags/env vars/config file options where appropriate

### Change log

- Timeouts use [ParseDuration](https://golang.org/pkg/time/#ParseDuration) to `time.Duration`s easier to input
- Add `--config` flag to specify a config input file
- Add `--nulldb` flag to use an in-memory database instead of couchbase
- Show example configuration file
- Manager, Docker, and Couchbase use default configurations
- Automatically add flags and env var bindings
  - Flags are lowercase config struct name hyphenated with fields. Ex: Docker.SaveDir -> --docker-save-dir
  - Env vars are uppercase config struct name underscored with fields, with SIGMA_ prefix. Ex: Docker.SaveDir -> SIGMA_DOCKER_SAVE_DIR

### TODO

Need tests


stephenrlouie:
side note, could you throw in some tests if you are going to reflection everything and implement a super generic way to do this

config/config.go
```diff
@@ -59,21 +70,169 @@ func GetConfig() (GlobalConfig, error) {
...
 	}
+	t = t.Elem()
+	if t.Kind() == reflect.Struct {
+		nf := t.NumField()
+		for i := 0; i < nf; i++ {
+			if t.Field(i).Kind() == reflect.Struct {
+				name := t.Type().Field(i).Name
 
-	return gCfg, nil
+				// viper.SetEnvPrefix("sigma-" + name)
```
stephenrlouie: prefix was moved to above right?

config/config.go
```diff
@@ -59,21 +70,169 @@ func GetConfig() (GlobalConfig, error) {
...
+func underscore(s string) string {
+	return splitCamel(s, '_')
+}
+
+// Underscore converts "CamelCasedString" to "camel-cased-string".
+func hyphen(s string) string {
+	return splitCamel(s, '-')
+}
+
+func splitCamel(s string, sep byte) string {
```
stephenrlouie: do you use these?

database/couchbase.go
```diff
@@ -36,43 +36,59 @@ var (
...
+	Timeout        string `desc:"Couchbase client timeout (ex: 1m30s)"`
+
+	timeout time.Duration
+	host    url.URL
 }
 
 //Couchbase defines an instance of a couchbase accessor
 type Couchbase struct {
-	cfg Config
+	cfg CouchbaseConfig
```
stephenrlouie: I think we can keep all the Configs just called Config, the real thing is these implementations of the interfaces should be in their own package. `database/couchbase` `task/docker` and so on. So when we ref it it would be `docker.Config` or `couchbase.Config`

database/couchbase.go
```diff
@@ -36,43 +36,59 @@ var (
...
 // Initialization of a DB connection
-func NewCouchbase(cfg Config) (*Couchbase, error) {
+func NewCouchbase(cfg CouchbaseConfig) (*Couchbase, error) {
 	var cb Couchbase
 	var err error
 	cb.cfg = cfg
-	cb.DB, err = gocb.Connect(cfg.Host.String())
 
-	log.Debug("Connected to couchbase at %s", cfg.Host.String())
+	if t, err := time.ParseDuration(cfg.Timeout); err != nil {
```
stephenrlouie: wow, `ParseDuration` is my hero

database/couchbase.go
adamschaub: In the current layout, the additional name is basically a requirement. Let's keep it this way until we decide we want to break out more packages for each different type of database/tasker/etc

config/config.go
adamschaub: yes

cmd/sigma.go
```diff
@@ -18,18 +18,18 @@ import (
 
 var SigmaCmd = &cobra.Command{
 	Use:   "sigma",
```
stephenrlouie: we are good with the command being ./bin/sigma sigma --config

cmd/sigma.go
```diff
@@ -62,14 +62,28 @@ func sigmaRun(cmd *cobra.Command, args []string) {
 	// Courtesy of patorjk -> http://patorjk.com/software/taag/#p=display&f=Graffiti&t=SiGmA
 
 	serviceGroup := service.New()
+
+	// TODO: Database
 	db := database.NewNullDB(false)
+	// 	var db database.Database
```
stephenrlouie: ?

stephenrlouie: why not just use couchbase

config.yaml
```diff
@@ -0,0 +1,8 @@
+---
```
stephenrlouie: provide the couchbase fields as if it was being deployed from [`./database/cbdeploy.sh`](https://github.com/sigma-dev/sigma/blob/master/database/cbdeploy.sh)

config/config.go
```diff
@@ -4,45 +4,58 @@ package config
...
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
+	"github.com/sigma-dev/sigma/database"
 	"github.com/sigma-dev/sigma/log"
+	"github.com/sigma-dev/sigma/manager"
 	"github.com/sigma-dev/sigma/task"
 )
 
+type Log struct {
```
stephenrlouie: ?

stephenrlouie: remove

config/config.go
```diff
@@ -58,21 +71,186 @@ func GetConfig() (GlobalConfig, error) {
...
+
+		// eachSubField only calls this function if both fieldName and subFieldName exist
+		field, _ := parent.Type().FieldByName(fieldName)
+		subField, _ := field.Type.FieldByName(subFieldName)
+		desc := subField.Tag.Get("desc")
+		switch subField.Type.Kind() {
+		case reflect.Bool:
+			sigmaCmd.PersistentFlags().Bool(flagStr, false, desc)
+		case reflect.Int:
+			fallthrough
```
meganokeefe: what is this `fallthrough` doing?

config/config.go
```diff
@@ -58,21 +71,186 @@ func GetConfig() (GlobalConfig, error) {
...
 
-	return gCfg, nil
+func flagString(parent, field string) string {
+	if len(parent) == 0 {
+		return strings.ToLower(hyphen(field))
+	}
+	if len(field) == 0 {
+		return strings.ToLower(hyphen(parent))
+	}
+	return strings.ToLower(hyphen(parent) + "-" + hyphen(field))
```
meganokeefe: can't you just have line `202` without the other ifs? what happens if you run `hyphen()` on an empty string

config/config.go
```diff
@@ -58,21 +71,186 @@ func GetConfig() (GlobalConfig, error) {
...
+}
+
+func envString(parent, field string) string {
+	if len(parent) == 0 {
+		return strings.ToUpper(underscore(field))
```
meganokeefe: see above comment on if these separate `if`s are needed?

config/config.go
```diff
@@ -58,21 +71,186 @@ func GetConfig() (GlobalConfig, error) {
...
+func isUpper(c byte) bool {
+	return c >= 'A' && c <= 'Z'
+}
+
+func isLower(c byte) bool {
+	return !isUpper(c)
+}
+
+func toUpper(c byte) byte {
+	return c - 32
```
meganokeefe: is this ascii conversion, have you tested that these helpers work

meganokeefe:
I see you have `tests` on your todo, but can you increase coverage a tad (currently 36%)? 
 `GetConfig()`  are `getConfig()` are currently not being tested - could add sanity tests for these?

manager/manager.go
```diff
@@ -16,9 +16,9 @@ import (
 const (
 	// Status of running operations should be queried at intervals
 	// no shorter than this value
-	DefaultMinOperationUpdateInterval = time.Second * 1
+	DefaultMinOperationUpdateInterval = "10s"
 
-	DefaultStatusInterval = time.Second * 20
+	DefaultStatusInterval = "20s"
 	DefaultRetryInterval  = time.Second * 20
```
meganokeefe: set `DefaultRetryInterval` to a string time, like the others?

manager/manager.go
```diff
@@ -46,27 +46,49 @@ func errorFilter(err error) error {
 }
 
 type Config struct {
+	OperationUpdateInterval string `desc:"Minimum duration between checking status of an operation"`
+	ServiceStatusInterval   string `desc:"Minimum duration between checking status of a service"`
```
meganokeefe: NIT, standardize the names of these config fields. both are poller-style update checkers.

maybe `ServiceOperationPollInterval` and `ServicePollInterval`? kinda long I guess

manager/manager_test.go
```diff
@@ -452,7 +458,7 @@ func TestUpdateServices(t *testing.T) {
 			test.svc.Set("Dirty", "yes")
 		}
 		if test.lastStatus {
-			test.svc.Set("LastStatus", time.Now().Add(-1*DefaultStatusInterval*2))
+			test.svc.Set("LastStatus", time.Now().Add(-2*m.serviceStatusInterval))
```
meganokeefe: what does this multiplying by -2 do??

adamschaub: Guaranteeing that the service meets the criteria for creating a Status operation. Making the `LastStatus` time more than one status interval in the past.

config/config.go
adamschaub: Everything except toUpper is caught in the tests, through testing `underscore` and `hyphen`

manager/manager.go
adamschaub: I don't see the need to include `poll` as part of the name. We can just rely on the desc~!!!!!!!!@*@&!&!*

adamschaub: SO THERE

manager/manager.go
adamschaub: Good point, don't know why I didn't add it as a configureable item.

config/config.go
stephenrlouie: Could we have the ASCII value and cast / type assert to a `byte`. Just for clarity

config.yaml
adamschaub: Done

config/config.go
adamschaub: It isn't, and I don't know why I did this. Fixed.

config/config.go
adamschaub: It's dumb, but sure

cmd/sigma.go
adamschaub: no

config/config.go
adamschaub: done

PR #33, opened 2017-05-01
Model definitions and DB Ops for Sites 

meganokeefe:
### PR checklist

- [x] Read the [contribution guidelines](https://www.google.com). TODO: fill this in with proper link
- [x] Filed the PR against the correct branch: master

### Change log

- Added Site definition to `model.go` 
- Added Site operations to `database.go` interface 
- Implemented Site DB operations in `couchbase.go`
- Updated nullDB 
- Regenerated mockDB 

### Notes

Couchbase tests don't exist yet, so holding off on testing. Should hopefully just be copypasta of the CRUD testing for the other DB types (Service, ServiceOp, etc.)


database/database.go
```diff
@@ -20,22 +20,27 @@ type Database interface {
...
 
 	UpdatePlugin(model.Plugin) error
 	UpdateService(model.Service) error
 	UpdateServiceOperation(model.ServiceOperation) error
+	UpdateSite(model.Site) error
 
 	QueryPlugins(model.Plugin) ([]model.Plugin, error)
 	QueryServices(model.Service) ([]model.Service, error)
 	QueryServiceOperations(model.ServiceOperation) ([]model.ServiceOperation, error)
+	QuerySites(model.Site) ([]model.Site, error)
```
stephenrlouie: can you implement the IDs one as well, that will return `[]string`

model/model.go
```diff
@@ -190,3 +190,16 @@ type ServiceOperation struct {
 
 	Service Service
 }
+
+//A Site encapsulates one or more native Task-Runners, and can be nested via Children
+type Site struct {
+	MetaImpl
+
+	Id string //autogenerated UUID
```
stephenrlouie: The API should generate these.

model/model.go
```diff
@@ -190,3 +190,16 @@ type ServiceOperation struct {
...
+
+	Name string //Can either be the ID of a native task runner (Physical), or a user-defined ID (Logical)
+
+	Physical bool //whether this site directly refers to a native Task Runner, or is a logical grouping of other Sites
+
+	Children []string //list of Site IDs
```
stephenrlouie: this will list the UUIDs right?

meganokeefe: This is a question- when a user defines a Site to group other Sites, should they list the internal sigma `Id`? (the uuid) Or should they list the `Name`. 

Ids are guaranteed unique. Sidenote, should we force unique site `Names?` Ie. you can only have one `edge1`?

stephenrlouie: I think they should internally be IDs but for the GUI they can be translated to human readable names, its the users fault to have duplicate names, but I guess you could.

stephenrlouie:
LGTM

PR #34, opened 2017-05-01
Sanity Unit Tests

stephenrlouie:
### Change log

- Just to get more test coverage, yes a lot of that is fairly simple code. Just wanted simple sanity tests.
- Better exclusions on the Makefile for testing


Makefile
```diff
@@ -2,7 +2,7 @@
 
 EXECUTABLE ?= sigma
 IMAGE ?= bin/$(EXECUTABLE)
-PACKAGES = $(shell go list ./... | grep -E -v '/bin|/vendor|/cmd|/docs|/api/v1/gen')
+PACKAGES = $(shell go list ./... | grep -E -v '/bin|/vendor|/cmd|/docs|/api/v1/gen|/api/v1/client|/database/mock_database|/task/mock_task|/plugin/examples')
```
adamschaub: What's the reason for this?

model/model_test.go
```diff
@@ -0,0 +1,137 @@
...
+func TestGet(t *testing.T) {
+	m := MetaImpl{}
+
+	// Get a nil key
+	a, ok := m.Get("a")
+	if ok {
+		t.Error("Did not return an error when expected")
+	}
+	if a != nil {
+		t.Error("Shold be nil")
```
adamschaub: Should

model/model_test.go
```diff
@@ -0,0 +1,137 @@
...
+func TestGetString(t *testing.T) {
+	m := MetaImpl{}
+
+	// GetString on a nil key
+	a, ok := m.GetString("a")
+	if ok {
+		t.Error("Did not return an error when expected")
+	}
+	if a != "" {
+		t.Error("Shold be nil")
```
adamschaub: NIT: Should

model/model_test.go
```diff
@@ -0,0 +1,137 @@
...
+	if !ok {
+		t.Error("Could not get key")
+	} else {
+		if a != 1 {
+			t.Error("Did not properly set key")
+		}
+	}
+}
+
+func TestGetString(t *testing.T) {
```
adamschaub: What happens when you `GetString` on a non-string value that has been previously added?

Makefile
stephenrlouie: to prevent those files from showing up when you run `make test`

model/model_test.go
stephenrlouie: tested now, good call!

Makefile
adamschaub: deal

adamschaub:
G2G

PR #35, opened 2017-05-02
Updated api spec for Site definitions

meganokeefe:
### PR checklist

- [x] Read the [contribution guidelines](https://www.google.com). TODO: fill this in with proper link
- [x] Filed the PR against the correct branch: master

### Change log

1.  New API spec, includes site definitions 
2. gitignore on `Gemfile.lock`, which is generated upon a local `jekyll serve`. 

### Notes

(A question: should we upgrade the API version at this point from v1 to v1.1, or not worth it?)


_docs/api-spec.md
```diff
@@ -184,10 +184,14 @@ font-style: italic;
     </style>
   </head>
   <body>
+  <h1>Mu-Sigma</h1>
```
stephenrlouie: why does this file have so much html, the others only have the markdown table at the top

stephenrlouie: mu?

PR #36, opened 2017-05-02
API support for Sites 

meganokeefe:
### Change log

- `model.go` -  Service and ServiceOp gets a Site. 
- `model.go` - `MarkedForDelete` and `Physical` are now `*bool` to support true-false database querying.
- updated API yaml and client 
- new sites routes 
- sites support for `convert.go`
- API server functions ^^ 
- `validateSite()` helper on Insert and Update of a site. (ensures no duplicates, ensures only logical sites are listed as parents, ensures no parent cycles) 
- added to `server_test.go` 
- shoehorned `couchbase_test.go` (site only testing) and excluded it in the makefile 
- `sitesGarbageCollect()` runs in `manager` with the ticker. sees sites `markedForDelete` and reassigns the parent of the site's children, prior to delete.


stephenrlouie:
can you remove all instances of `mu` from this. I think swagger is generating it?

api/v1/client/plugin.go
```diff
@@ -1,16 +1,17 @@
...
+ * This outlines version 1 of the Sigma API. All requests should be prefixed with the `v1` basePath. 
  *
  * OpenAPI spec version: 1.0
- *
+ * 
  * Generated by: https://github.com/swagger-api/swagger-codegen.git
  */
 
-package apiclient
+package swagger
```
stephenrlouie: hesitant this changes, want to be copy pastable when we need updates, should use the Makefile

api/v1/client/plugin_io.go
```diff
@@ -1,19 +1,20 @@
...
- * Mu-Sigma
+/* 
+ * Sigma
  *
- * This outlines version 1 of the Mu Sigma API. All requests should be prefixed with the `v1` basePath.
+ * This outlines version 1 of the Sigma API. All requests should be prefixed with the `v1` basePath. 
  *
  * OpenAPI spec version: 1.0
- *
+ *
```
stephenrlouie: remove the spaces

api/v1/client/plugin_io.go
```diff
@@ -1,19 +1,20 @@
-/*
- * Mu-Sigma
+/*
```
stephenrlouie: remove space

api/v1/server/server.go
```diff
@@ -121,6 +121,31 @@ func (s *Server) AddService(w http.ResponseWriter, r *http.Request) {
...
+func (s *Server) AddSite(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+	decoder := json.NewDecoder(r.Body)
+	defer r.Body.Close()
+	var apiObj apiclient.Site
+	var err error
+	if err = decoder.Decode(&apiObj); err != nil {
+		log.Errorf("Decoder error for API-derived site: %v", err)
+	}
+	if apiObj.Name == "" || hasInvalidNames(apiObj) {
```
stephenrlouie: manager doesn't do any split on `.` so hasInvalidNames should not have to be called.

api/v1/server/server.go
```diff
@@ -121,6 +121,31 @@ func (s *Server) AddService(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusCreated)
 }
 
+//AddSite implements the handler to process the JSON site and insert it into the DB
+func (s *Server) AddSite(w http.ResponseWriter, r *http.Request) {
```
stephenrlouie: need to generate a UUID.

api/v1/gen/v1.yaml
```diff
@@ -282,3 +359,17 @@ definitions:
           type: object
         from:
           type: object
+
+  Site:
+      type: object
+      properties:
+        id:
```
stephenrlouie: they users shouldn't submit this. but could see it on a return

api/v1/server/server.go
stephenrlouie: (all the Add handlers should generate a UUID)

meganokeefe: The UUID stuff is currently happening in `couchbase.go`. Was just following convention https://github.com/sigma-dev/sigma/blob/master/database/couchbase.go#L255

But I can move all the UUID stuff to `api/server` if ya like.

api/v1/gen/v1.yaml
meganokeefe: was just copying the convention of Service and Plugin in the swagger spec

https://github.com/sigma-dev/sigma/blob/master/api/v1/gen/v1.yaml#L227

Should IDs be removed from these as well?

api/v1/client/default_api.go
```diff
@@ -159,6 +159,66 @@ func (a DefaultApi) AddService(body Service) (*APIResponse, error) {
...
+
+	// set Content-Type header
+	localVarHttpContentType := a.Configuration.APIClient.SelectHeaderContentType(localVarHttpContentTypes)
+	if localVarHttpContentType != "" {
+		localVarHeaderParams["Content-Type"] = localVarHttpContentType
+	}
+	// to determine the Accept header
+	localVarHttpHeaderAccepts := []string{
+		"application/json",
+		}
```
stephenrlouie: is this linted?

api/v1/convert/convert.go
```diff
@@ -33,6 +33,15 @@ func FromModelService(o model.Service) apiclient.Service {
 	}
 }
 
+func FromModelSite(o model.Site) apiclient.Site {
```
stephenrlouie: could you implement a simple test function for this

api/v1/gen/v1.yaml
stephenrlouie: it works. I was making a note, not a change request

api/v1/server/server.go
stephenrlouie: sounds fine leaving it to the DB, as long as its clear that any other DB implementations are responsible for it

api/v1/server/server.go
```diff
@@ -121,6 +121,79 @@ func (s *Server) AddService(w http.ResponseWriter, r *http.Request) {
...
+//validateSite is a helper that validates the Name, type, and Children of a site before DB insertion
+func (s *Server) validateSite(site apiclient.Site) error {
+	if site.Name == "" || hasInvalidNames(site) {
+		return fmt.Errorf("site has invalid or empty name (%s)", site.Name)
+	}
+	//physical sites must have 0 children and vice versa
+	if site.Physical == true && len(site.Children) != 0 {
+		return fmt.Errorf(ErrPhysicalSiteHasChildren)
+	}
+	if len(site.Children) == 0 && site.Physical == false {
```
stephenrlouie: is it necessary to be this strict? I'm just imagining being in a UI and dragging children sites under my logical sites. Could see myself clicking and dragging children around and leaving a parent childless for a time even though thats not my final state.

api/v1/server/server.go
```diff
@@ -121,6 +121,79 @@ func (s *Server) AddService(w http.ResponseWriter, r *http.Request) {
...
+
+	for _, c := range site.Children {
+		childSites, err := s.db.QuerySites(model.Site{Name: c})
+		if err != nil {
+			return fmt.Errorf(ErrValidationQuery)
+		}
+		if len(childSites) == 0 {
+			return fmt.Errorf(ErrChildNotExists)
+		}
+		//TODO - find a better way to check if each child is parentless. this is awful. worth adding a Parent field to Site?
```
stephenrlouie: i dont think they need to have a logical parent.

stephenrlouie: you should get all sites, then query sites that have `Child: Id` and ensure the length of returns is not greater than 1. it could be 0. 

NIT: you could also do ID only operations to save sending the entire structs around

api/v1/server/server.go
```diff
@@ -153,6 +226,22 @@ func (s *Server) DeleteService(w http.ResponseWriter, r *http.Request) {
...
+func (s *Server) DeleteSite(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+	vars := mux.Vars(r)
+	id, ok := vars["site-id"]
+	if !ok || id == "" {
+		http.Error(w, "No Site ID Provided.", http.StatusBadRequest)
+		return
+	}
+	if err := s.db.DeleteSite(id); err != nil {
+		http.Error(w, err.Error(), http.StatusNotFound)
```
stephenrlouie: don't think 404 is right since this endpoint is functional and found. but internally we cannot find its related DB entry

api/v1/server/server.go
```diff
@@ -276,6 +365,53 @@ func (s *Server) GetServices(w http.ResponseWriter, r *http.Request) {
...
+	vars := mux.Vars(r)
+	id, ok := vars["site-id"]
+	if !ok || id == "" {
+		http.Error(w, "No Site ID Provided.", http.StatusBadRequest)
+		return
+	}
+	var modObj model.Site
+	var err error
+	if modObj, err = s.db.GetSite(id); err != nil {
+		http.Error(w, "Site Not Found.", http.StatusNotFound)
```
stephenrlouie: same as above, 404?

api/v1/server/server.go
```diff
@@ -392,6 +568,11 @@ func hasInvalidNames(inf interface{}) bool {
 				return true
 			}
 		}
+	case apiclient.Site:
```
stephenrlouie: shouldn't have to filter out `.` from sites. Manager is not using it like it does `<Plugin>.<Service>.Input`

api/v1/server/server_test.go
```diff
@@ -167,36 +258,73 @@ func TestUpdate(t *testing.T) {
...
 	}
-	if v, _ := client.DeleteService("fullService"); v.StatusCode != 200 {
-		t.Log("Expected 200, got ", v.StatusCode)
-	}
 	ps, _, _ := client.GetPlugins()
-	ss, _, _ := client.GetServices("")
 	if len(ps) != (len(plugins) - 1) {
 		t.Errorf("Plugins returned not equal. len() = %v, should be %v", len(ps), len(plugins)-1)
 	}
+
```
stephenrlouie: same as above, one that exists and one that does not

api/v1/client/default_api.go
meganokeefe: it wasn't, no. worth adding a `golinter` command to the `Makefile`, @mikeynap ?

api/v1/server/server.go
meganokeefe: Good point. This is a philosophical question, then. Like how do we know when the state is the "final state?" For instance in the UI. Not sure. 

One option is to leave this code intact, but in the UI, let the user click and drag things around, but only press `GO` when they're ready to update their configuration for real. At which point the API calls will happen. 

This also reminded me to add validate logic to `UpdateSite` as well.

api/v1/server/server.go
meganokeefe: ? https://stackoverflow.com/questions/5604816/whats-the-most-appropriate-http-status-code-for-an-item-not-found-error-page

stephenrlouie:
remove `gen/` files

api/v1/convert/convert.go
```diff
@@ -76,6 +85,15 @@ func FromAPIService(o apiclient.Service) model.Service {
 	return svc
 }
 
+func FromAPISite(o apiclient.Site) model.Site {
```
stephenrlouie: should also test this, some bi-directional test like the other convert tests.

database/database.go
```diff
@@ -7,6 +7,15 @@ import (
 	"github.com/sigma-dev/sigma/model"
 )
 
+const (
```
stephenrlouie: maybe these belong in model? the more I think about it, the more I think DB should simply put things in and out, not concerned with contents

database/database.go
```diff
@@ -7,6 +7,15 @@ import (
...
+	pluginType           = "PLUGIN"
+	serviceType          = "SERVICE"
+	serviceOperationType = "SERVICE_OPERATION"
+	siteType             = "SITE"
+	Logical              = "LOGICAL"
```
stephenrlouie: I think these consts belong somewhere else. Since we don't have separate functions in db to handle a logical or physical, these should be defined by the person creating and managing the Sites construct. so probably API?

api/v1/server/server.go
```diff
@@ -359,6 +515,93 @@ func (s *Server) UpdateService(w http.ResponseWriter, r *http.Request) {
...
+	// Sites cannot have duplicate names
+	duplicateNames, err := s.db.QuerySites(model.Site{Name: site.Name})
+	if err != nil {
+		return fmt.Errorf("err querying for dup. sites where name=%s: %v", site.Name, err)
+	}
+	if len(duplicateNames) > 0 {
+		return fmt.Errorf(ErrDuplicateName)
+	}
+	// Type check - must be physical or logical
+	if !((site.Type == database.Physical) || (site.Type == database.Logical)) {
```
stephenrlouie: is this someone elses problem if the `consts` move? maybe.

api/v1/server/server.go
```diff
@@ -153,11 +187,85 @@ func (s *Server) DeleteService(w http.ResponseWriter, r *http.Request) {
...
+func (s *Server) DeleteSite(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+	vars := mux.Vars(r)
+	name, ok := vars["site-name"]
+	if !ok || name == "" {
+		http.Error(w, "No Site name Provided.", http.StatusBadRequest)
+		return
+	}
+	// First, find the site they're referring to
+	sites, err := s.db.QuerySites(model.Site{Name: name})
```
stephenrlouie: why not GetSite

stephenrlouie: why not Get?

api/v1/server/server.go
```diff
@@ -153,11 +187,85 @@ func (s *Server) DeleteService(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, fmt.Sprintf("No site found with name=%s", name), http.StatusNotFound)
+		return
+	}
+	if len(sites) > 1 {
+		http.Error(w, err.Error(), http.StatusNonAuthoritativeInfo)
+		return
+	}
+	// Do the Delete
+	siteForDeletion := sites[0]
+	if err := s.db.DeleteSite(siteForDeletion.Id); err != nil {
```
stephenrlouie: dangerous because we can orphan people if this deletes and then the thread crashes. should resolve first? Then we can implement logic to handle duplicates and realize if there are two parents, and one of them lists the other as a parent, we meant to delete that.

api/v1/server/server.go
```diff
@@ -153,11 +187,85 @@ func (s *Server) DeleteService(w http.ResponseWriter, r *http.Request) {
...
+			log.Error(err)
+			return
+		}
+	}
+	w.WriteHeader(http.StatusOK)
+}
+
+// Given a site name, resolve the parentless children
+// This function only has to be called on logical sites (deleting phys sites has no consequences)
+func (s *Server) deleteSiteResolution(myself model.Site) error {
```
stephenrlouie: what if this dies without re-assigning everything? Maybe a period cleanup is necessary

api/v1/server/server.go
```diff
@@ -359,6 +515,93 @@ func (s *Server) UpdateService(w http.ResponseWriter, r *http.Request) {
...
+
+	updateFunc := func() (int, error) {
+		var site2Update model.Site
+		var err error
+		if site2Update, err = s.db.GetSite(id); err != nil {
+			return http.StatusNotFound, err
+		}
+		updatedSite := convert.FromAPISite(apiObj)
+		updatedSite.MetaImpl = site2Update.MetaImpl
+		updatedSite.Set("Dirty", "true")
```
stephenrlouie: what is this for?

api/v1/server/server_test.go
```diff
@@ -48,8 +51,71 @@ func TestMain(m *testing.M) {
...
+		},
+		{ // GOOD - parentless physical site
+			apiclient.Site{Name: "LA", Type: database.Physical},
+			nil,
+		},
+		{ // GOOD - parentless, childless logical site
+			apiclient.Site{Name: "ROGUEONE", Type: database.Logical},
+			nil,
+		},
+		{ // GOOD - a nonexistent parent should get added
```
stephenrlouie: existing Parent that is logical test

stephenrlouie: Cycle test

api/v1/server/server_test.go
```diff
@@ -165,36 +262,182 @@ func TestUpdate(t *testing.T) {
...
+		//now GetSite to ensure it updated
+		postUpdate, err := db.GetSite(maSites[0].Id)
+		if err != nil {
+			t.Errorf("Get err after updating: %v", err)
+		}
+		if postUpdate.Name != "MASS-NEW-NAME" {
+			t.Errorf("Site update did not succeed in DB- its name should be MASS-NEW-NAME, actual is: %s", postUpdate.Name)
+		}
+	}
+	//db.Clear()
```

api/v1/server/server_test.go
```diff
@@ -165,36 +262,182 @@ func TestUpdate(t *testing.T) {
 		t.Errorf("Update Value differs. %+v vs %+v", ep.Input, p.Input)
 	}
 	if v, _ := client.UpdatePlugin("SDFEIFIFIFI", ep); v.StatusCode != 404 {
-		t.Error("Should have returned 404, got:", v.StatusCode)
+		t.Error("Plugin Update, Should have returned 404, got:", v.StatusCode)
 	}
 
+	//service. TODO - add a passing test here
```
stephenrlouie: remove?

api/v1/server/server_test.go
```diff
@@ -165,36 +262,182 @@ func TestUpdate(t *testing.T) {
...
+	//site - good (update an existing site, then check for success)
+	maSites, err := db.QuerySites(model.Site{Name: "MASS"})
+	if err != nil {
+		t.Errorf("Sites query err: %v", err)
+	}
+	if len(maSites) != 1 {
+		t.Errorf("Pre-update nullDB query for sites failed, expected len=1, got len=%d", len(maSites))
+	} else {
+		updatedSite := apiclient.Site{Id: maSites[0].Id, Name: "MASS-NEW-NAME", Type: database.Logical}
+		//t.Logf("About to update with site: %+v", updatedSite)
```

api/v1/server/server_test.go
```diff
@@ -165,36 +262,182 @@ func TestUpdate(t *testing.T) {
...
+	east := apiclient.Site{Name: "east", Type: database.Logical, Parent: "usa"}
+	nyc := apiclient.Site{Name: "nyc", Type: database.Physical, Parent: "east"}
+	west := apiclient.Site{Name: "west", Type: database.Logical, Parent: "usa"}
+	ca := apiclient.Site{Name: "ca", Type: database.Logical, Parent: "west"}
+	sfo := apiclient.Site{Name: "sfo", Type: database.Physical, Parent: "ca"}
+	la := apiclient.Site{Name: "la", Type: database.Physical, Parent: "ca"}
+	seattle := apiclient.Site{Name: "seattle", Type: database.Logical, Parent: "west"}
+
+	//db.Clear()
+	client.AddSite(usa)
```
stephenrlouie: should make sure these work. and t.Fatal or something if they don't

api/v1/convert/convert.go
stephenrlouie: tests?

api/v1/server/server.go
stephenrlouie: should we support both by ID and name like the other queries. If so we need to attempt a `IdFromName` conversion

api/v1/server/server.go
```diff
@@ -359,6 +492,135 @@ func (s *Server) UpdateService(w http.ResponseWriter, r *http.Request) {
...
+		log.Errorf("Update err: %v", err)
+		http.Error(w, err.Error(), rc)
+		return
+	}
+}
+
+//validateSite is a helper that validates the Name and Parent before Add and Update
+func (s *Server) validateSite(site apiclient.Site) error {
+	// Sites cannot have duplicate names
+	duplicateNames, err := s.db.QuerySites(model.Site{Name: site.Name})
```
stephenrlouie: are we required to have name?

api/v1/server/server.go
```diff
@@ -359,6 +492,135 @@ func (s *Server) UpdateService(w http.ResponseWriter, r *http.Request) {
...
+		myParent, err := s.db.QuerySites(model.Site{Name: site.Parent})
+		if err != nil {
+			return fmt.Errorf("err querying children for site=%s: %v", site.Name, err)
+		}
+		if len(myParent) == 0 {
+			_, err := s.db.InsertSite(model.Site{Name: site.Parent, Kind: database.Logical})
+			if err != nil {
+				return err
+			}
+			temp, err := s.db.QuerySites(model.Site{Name: site.Parent})
```
stephenrlouie: why query after insert?

meganokeefe: so that we can get the whole Parent struct with the ID and stuff. I pass a parent/`model.Site` into the cycle detection stuff.

Idk I didn't want to pass around a hardcoded parent that I created right then. Wanted the actual DB struct I guess.

api/v1/server/server.go
meganokeefe: I can see the usefulness of this, yeah. I'm doing a fair amount of querying.

IDK - up to you if we should add it.

database/database.go
```diff
@@ -7,11 +7,22 @@ import (
 	"github.com/sigma-dev/sigma/model"
 )
 
+const (
+	pluginType           = "PLUGIN"
+	serviceType          = "SERVICE"
+	serviceOperationType = "SERVICE_OPERATION"
+	siteType             = "SITE"
+	Logical              = "Logical"
```
stephenrlouie: I think these should live somewhere else. If anyone wants to use sites and have access to these consts, they have to `import db`

manager/manager.go
```diff
@@ -130,6 +134,54 @@ func (m *manager) Stop() {
...
+		// get all my children
+		children, err := m.db.QuerySites(model.Site{Parent: myself.Name})
+		if err != nil {
+			log.WithError(err)
+			continue
+		}
+		// resolve children, if I have them
+		if len(children) > 0 {
+			var newDaddy string
+			if myself.Parent != "" {
```
stephenrlouie: do we need this, wouldn't `newDaddy = myself.Parent` always work?

model/model.go
```diff
@@ -208,7 +212,9 @@ type Site struct {
 
 	Name string //Can either be the ID of a native task runner (Physical), or a user-defined ID (Logical)
 
-	Physical bool //whether this site directly refers to a native Task Runner, or is a logical grouping of other Sites
+	Kind string //logical or physical
+
+	Parent string //Site.name
 
-	Children []string //list of Site IDs
+	MarkedForDelete bool //whether this site is marked for deletion
```
stephenrlouie: cannot be a bool, need to make an ENUM to distinguish between a "zero value", and false and true when someone makes a site query. We won't be able to distinguish between I want all sites that have `MarkedForDelete==false`, and someone who doesn't care what the value is.

api/v1/convert/convert_test.go
```diff
@@ -8,6 +8,37 @@ import (
 	"github.com/sigma-dev/sigma/model"
 )
 
+var (
+	T = true
+	F = false
+)
+
+// sanity test for fromModelSite
+func TestFromModelSite(t *testing.T) {
```
stephenrlouie: merge like the other `convertTests`

api/v1/server/server.go
stephenrlouie: shoot my B, we already support this. https://github.com/sigma-dev/sigma/blob/sites-api/database/couchbase.go#L337 which will then try by ID and name

stephenrlouie: I think we should support both since other deletes will [DeleteById](https://github.com/sigma-dev/sigma/blob/sites-api/database/couchbase.go#L186-L200)

api/v1/client/default_api.go
mikeynap: Adding it for the generator seems like a good call.

stephenrlouie:
LGTM

PR #39, opened 2017-05-03
Site processing for manager.newOperation

meganokeefe:
### Change log

- [x] added site processing to `manager.newOperation()`, with the `choosePhysicalSite()` helper for logical sites. (picks a physical child at random.)
- [x] testing for site processing (and 'functionally' tested `choosePhysicalSite()` by adding a service via the API that used a logical site w/ multiple children -- a new site was chosen each time for the `ServiceOperation`. 
- [x] added a test for `choosePhysicalSite` to the manager unit tests


manager/manager.go
```diff
@@ -71,6 +75,11 @@ func NewManager(db database.Database, tasker task.Tasker, cfg Config) *manager {
 
 // Start implements service to begin the manager
 func (m *manager) Start() error {
+	// On start, insert default site=local. TODO: is this the right place to add local?
```
stephenrlouie: i think this is fine

manager/manager.go
```diff
@@ -102,6 +106,11 @@ func NewManager(db database.Database, tasker task.Tasker, cfg Config) (*manager,
...
+	_, err := m.db.InsertSite(model.Site{Name: "local", Type: database.Physical, Parent: ""})
```
adamschaub: If the local site already exists, insert will report an error, right? Should probably filter out the "KeyExists" error.

manager/manager.go
```diff
@@ -591,6 +623,52 @@ func (m *manager) newOperation(svc *model.Service, plugin model.Plugin, opType s
...
+	if err != nil {
+		return "", err
+	}
+	if len(phys) == 0 {
+		return "", ErrEmptyLogicalSite
+	}
+	// choose physical site at random
+	rand.Seed(time.Now().Unix())
+	index := rand.Intn(len(phys) - 1)
+	chosenOne := phys[index]
```
adamschaub: Can we call this `neo`

manager/manager.go
```diff
@@ -104,6 +108,13 @@ func NewManager(db database.Database, tasker task.Tasker, cfg Config) (*manager,
 
 // Start implements service to begin the manager
 func (m *manager) Start() error {
+	// On start, insert default site=local, only if it's not already there
+	if _, err := m.db.GetSite("local"); err != nil {
+		_, insertErr := m.db.InsertSite(model.Site{Name: "local", Physical: &T, Parent: ""})
+		if insertErr != nil {
+			log.WithError(insertErr).Error("Failed to insert local site on manager startup")
```
adamschaub: Can we operate at all without the default site? Should this be cause to shutdown?

manager/manager.go
```diff
@@ -104,6 +108,13 @@ func NewManager(db database.Database, tasker task.Tasker, cfg Config) (*manager,
 
 // Start implements service to begin the manager
 func (m *manager) Start() error {
+	// On start, insert default site=local, only if it's not already there
+	if _, err := m.db.GetSite("local"); err != nil {
```
adamschaub: Differentiate generic errors and `database.ErrNotFound`. It could be the database configuration is off, and not that the site doesn't exist, in which case start quit.

manager/manager.go
meganokeefe: Probably, yes.

manager/manager.go
```diff
@@ -573,19 +581,37 @@ func (m *manager) inputSatisfied(svc *model.Service, plugin model.Plugin) (satis
 
 // newOperation attempts to create a new ServiceOperation with the provided service.
 func (m *manager) newOperation(svc *model.Service, plugin model.Plugin, opType string) error {
+	// Process the service site. if Physical, use it. if Logical, pick a child.
+	chosenOne := svc.Site
+	givenSite, err := m.db.GetSite(svc.Site)
+	if err != nil {
+		return fmt.Errorf("Site=%s provided for service=%s not found: %v", svc.Site, svc.Name, err)
```
adamschaub: `ErrSiteNotFound`?

manager/manager.go
```diff
@@ -573,19 +581,37 @@ func (m *manager) inputSatisfied(svc *model.Service, plugin model.Plugin) (satis
...
+	}
+	// if a logical site is provided, choose a random physical child site, and use that instead for the Op
+	if *givenSite.Physical == false {
```
adamschaub: Any chance this might be `nil`? I'll circle back after I've read all the code if this is the case. Otherwise, we have a potential panic situation.

manager/manager.go
```diff
@@ -573,19 +581,37 @@ func (m *manager) inputSatisfied(svc *model.Service, plugin model.Plugin) (satis
...
 	// Create operation, if one isn't currently active
 	opId := fmt.Sprintf("%s-operation", svc.Id)
-	if _, err := m.db.GetServiceOperation(opId); err != database.ErrKeyNotFound {
+	// we expect a keyNotFound err
+	if _, err := m.db.GetServiceOperation(opId); err != nil && err.Error() != database.ErrKeyNotFound.Error() {
 		if err == nil {
 			return ErrOperationExists
 		}
-		return err
+		return fmt.Errorf("Err getting ServiceOperation for id=%s: %v", opId, err)
```
adamschaub: why wrap in another error?

manager/manager.go
```diff
@@ -594,14 +620,59 @@ func (m *manager) newOperation(svc *model.Service, plugin model.Plugin, opType s
 	// No plugin update
 	// Save operation
 	if op, err := m.db.InsertServiceOperation(op); err != nil {
-		return err
+		return fmt.Errorf("InsertServiceOperation error: %v", err)
```
adamschaub: same as above

adamschaub: If we want more insight, I might recommend we log and then pass the original error up. Being that manager might react to different errors discretely.

manager/manager.go
```diff
@@ -594,14 +620,59 @@ func (m *manager) newOperation(svc *model.Service, plugin model.Plugin, opType s
...
+func (m *manager) choosePhysicalSite(logicalRoot string) (string, error) {
+	phys, err := m.getAllPhysicalSites(logicalRoot)
+	if err != nil {
+		return "", err
+	}
+	if len(phys) == 0 {
+		return "", ErrEmptyLogicalSite
+	}
+	// choose physical site at random
+	rand.Seed(time.Now().Unix())
```
adamschaub: Seed every time, or do we just seed during start up?

adamschaub: Not a huge deal I guess

adamschaub: So NIT

manager/manager.go
```diff
@@ -594,14 +620,59 @@ func (m *manager) newOperation(svc *model.Service, plugin model.Plugin, opType s
...
+func (m *manager) getAllPhysicalSites(parentName string) ([]string, error) {
+	allPhysicalSites := []string{}
+
+	// get all my children, both physical and logical
+	children, err := m.db.QuerySites(model.Site{Parent: parentName})
+	if err != nil {
+		return nil, err
+	}
+
+	//log.Infof("Parent is %s, I got all children: %+v", parentName, children)
```
adamschaub: Remove

manager/manager.go
```diff
@@ -594,14 +620,59 @@ func (m *manager) newOperation(svc *model.Service, plugin model.Plugin, opType s
...
+
+	for _, child := range children {
+		//if physical, base case -> add me
+		if *child.Physical == true {
```
adamschaub: Maybe nil?

manager/manager.go
meganokeefe: Yes you're right -- adding a nil check to every `.Physical` reference.

manager/manager.go
meganokeefe: Fixed

manager/manager.go
```diff
@@ -594,14 +620,59 @@ func (m *manager) newOperation(svc *model.Service, plugin model.Plugin, opType s
...
 	} else {
 		log.Debugf("Service %s created %s operation", svc.Name, opType, op)
 	}
-
 	return nil
 }
 
+// choosePhysicalSite gets every Physical site under a given Logical site, then chooses one at random
+// returns the name of the random site and an error if there are no Physical sites under the logicalRoot
+func (m *manager) choosePhysicalSite(logicalRoot string) (string, error) {
```
stephenrlouie: `LetsGetPhysical` please

manager/manager.go
```diff
@@ -104,6 +110,19 @@ func NewManager(db database.Database, tasker task.Tasker, cfg Config) (*manager,
 
 // Start implements service to begin the manager
 func (m *manager) Start() error {
+	// On start, insert default site=local, only if it's not already there
+	if _, err := m.db.GetSite("local"); err != nil {
+		if err == ErrSiteNotFound {
```
stephenrlouie: needs to be the `db.ErrNotFound` error since DB will never return the err declared here

manager/manager.go
adamschaub: yes please

stephenrlouie:
We should reject services that do not have an existing site.

meganokeefe:
@stephenrlouie i'm assuming you mean reject incoming services from the API (in add and update) that have no site listed. 

So that's the change i just pushed.

adamschaub:
LGTM. I'll merge later, once the time casting fixes land. Just to do a last e2e test.

PR #41, opened 2017-05-03
adding a version sub command

stephenrlouie:
### Change log

Adding a version command 

```
$./bin/sigma version
Version: 0.1
```


cmd/version.go
```diff
@@ -0,0 +1,20 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+const version = "0.1"
```
meganokeefe: what's going to happen when we have a new version? is this hardcoded

meganokeefe:
LGTM except for that one question.

cmd/version.go
stephenrlouie: yea someone will have to change this when we have tagged releases

PR #42, opened 2017-05-03
task integration test 

stephenrlouie:
### Change log

task test was an integration test since it leverages an external dependency aka: docker. 
- Set up a separate directory structure for integration tests that will only run daily. 
- Modified the makefile to "run" setup and shutdown scripts for when real set up is required. 
- Modified the makefile to have a directive `make integration`

in other words:
> A PR into the Sigma-dev Github repository for the purpose of moving tests from the task subpackage into another specified testing directory for the affect of removing docker as a prerequisite for running unit tests - @adamschaub


meganokeefe:
Nice, LGTM. 

Since you put in the PR templates, I should probably suggest that you add `TODOs` to write the integration startup/shutdown scripts. And expand out anything else you've stubbed here.

stephenrlouie:
TODOs not required here. Will be required for the DB tests since it will stand up a instance of couchbase. So its up to @vgunning when she implements the tests against CB.

PR #49, opened 2017-05-05
configuring server, url.Parse function

stephenrlouie:
### Change log

- Adds Server configuration of `Server.URL`
- Also added `f` to some formatted debug log statements in couchbase.
### Notes

Uses the [URL.Parse](https://golang.org/pkg/net/url/#Parse) function which expects a PORT. when we use a FQDN we may have to change this to be more flexible.


cmd/sigma.go
```diff
@@ -85,7 +85,13 @@ func sigmaRun(cmd *cobra.Command, args []string) {
 	}
 
 	// Sigma API Server
-	serviceGroup.Add(server.New(":2481", db))
+	s, err := server.New(gCfg.Server, db)
```
adamschaub: Do this like the rest
```
var server ServerType
if server, globalErr = servier.New(...); globalErr != nil {
  return
}
...
```

meganokeefe:
V nice, LGTM.

PR #51, opened 2017-05-05
Docker unit tests

adamschaub:


task/docker_test.go
```diff
@@ -0,0 +1,480 @@
...
+			if err != nil && len(test.errContains) != 0 && strings.Contains(err.Error(), test.errContains) {
+				continue
+			}
+			t.Errorf("Got err %v, expected %v", err, test.err)
+		}
+	}
+}
+
+func TestDockerPull(t *testing.T) {
+	// - client.ImagePull <- resp, err := cli.post(ctx, "/images/create", query, nil, headers)
```
stephenrlouie: ?

adamschaub: WHOOPS. That was me cataloguing the client calls that would be made.

task/docker_test.go
```diff
@@ -0,0 +1,494 @@
...
+					Name:  "hi",
+					Type:  "string",
+					Value: []string{"127.0.0.1"},
+				},
+			},
+		},
+	}
+}
+
+var saveDir DockerConfig
```
meganokeefe: nit, `saveDir` implies a string type, I feel, not a Config `struct`

PR #52, opened 2017-05-08
Integration Tests for the Couchbase DB Interface

vgunning:
### PR checklist

- [X] Read the [contribution guidelines](https://www.google.com). TODO: fill this in with proper link
- [X] Filed the PR against the correct branch: master

### Change log

Unit tests for all of the exported functions.
Tests are table tests for ease of expansion.
Tests include messages for tracking a failing test (easier than matching a number).


tests/integration/database/couchbase_test.go
```diff
@@ -0,0 +1,881 @@
...
+		"key2": "value2",
+	}
+	testPlugin = model.Plugin{
+		Name:        "test",
+		Summary:     "plugin couchbase test",
+		Description: "this plugin is a couchbase test object",
+		Maintainer:  "parrotfish",
+	}
+	testSvc = model.Service{
+		Name:          "test",
```
meganokeefe: nit, can the plugin and service have different names?

meganokeefe:
`couchbase.go` at 81% test coverage, nice going!

tests/integration/database/couchbase/couchbase_test.go
```diff
@@ -0,0 +1,901 @@
...
+		emptyMap := make(map[string]interface{})
+		_, err = cb.BKT.Get(id, &emptyMap)
+		if err != gocb.ErrKeyNotFound {
+			t.Fatalf("Clear Test %d: Failed.\nFailed to clear Couchbase.\n Error: %v", i, err)
+		}
+	}
+}
+
+// Test getting an ID from the Name
+func TestIdFromName(t *testing.T) {
```
meganokeefe: Tests where you have more than 3 or 4 cases, might be more idiomatic to put into tabletests. Particularly in a test like `TestIdFromName` where you're checking the result against the same variable (`id`).

tests/integration/database/couchbase/couchbase_test.go
```diff
@@ -0,0 +1,901 @@
...
+				i, err, test.expected)
+		} else {
+			t.Logf("InsertServiceOperation Test %d: Passed.", i)
+		}
+		cb.Clear()
+	}
+}
+
+// Test the update plugin
+func TestUpdatePlugin(t *testing.T) {
```
meganokeefe: all of these Update tests have only one (passing) test case. I'd at the least add one failing case to each Update test, and ideally make these tabletests.

meganokeefe: nit, make these updateTests table tests -- reduces space, makes the tests easier to read and add to later.

tests/integration/database/couchbase/couchbase_test.go
```diff
@@ -0,0 +1,901 @@
...
+	if !pluginsAreEqual(cbPlugin, updatePlugin) {
+		t.Error("UpdatePlugin Test Failed.\nUpdated object doesn't match.")
+	}
+	updateID := cb.IdFromName(updatePlugin.Name, db.PluginType)
+	if updatePlugin.Id != updateID {
+		t.Errorf("UpdatePlugin Test Failed.\nUpdated name failed to match ID.\nID from Name: %v\nID: %v",
+			updateID, id)
+	}
+	oldID := cb.IdFromName(testPlugin.Name, db.PluginType)
+	t.Logf("ID from old name: %v", oldID)
```
meganokeefe: remove

tests/integration/database/couchbase/couchbase_test.go
```diff
@@ -0,0 +1,901 @@
...
+
+	match := testPlugin
+	match.Name = ""
+	match.Id = ""
+	cbPlugins, err = cb.QueryPlugins(match)
+	if len(cbPlugins) != 1 {
+		t.Errorf("QueryPlugins Test Failed.\nUnexpected error on mismatched query.\nError: %v", err)
+	} else if !pluginsAreEqual(testPlugin, cbPlugins[0]) {
+		t.Error("QueryPlugins Test Failed.\nMutated object.")
+	}
```
meganokeefe: clear the DB in between each test case. 

and ideally make this a tabletest

tests/integration/database/couchbase/couchbase_test.go
meganokeefe: don't have to adhere to my syntax (i'll change my tests to match yours), but here's how I did `TestUpdateSite`: https://github.com/sigma-dev/sigma/blob/sites-api/database/couchbase_test.go#L103

tests/integration/database/couchbase/couchbase_test.go
```diff
@@ -0,0 +1,901 @@
...
+	if len(cbPlugins) > 0 {
+		t.Errorf("QueryPlugins Test Failed.\nUnexpected match on query.\nError: %v", err)
+		t.Logf("Query: %+v", mismatched)
+		t.Logf("Matched: %+v", cbPlugins)
+	}
+	cb.Clear()
+}
+
+// Test the query services
+func TestQueryServices(t *testing.T) {
```
meganokeefe: make query tests table tests

tests/integration/database/couchbase/couchbase_test.go
```diff
@@ -0,0 +1,901 @@
...
+	if len(cbPluginsIds) > 0 {
+		t.Errorf("QueryPluginsIds Test Failed.\nUnexpected match on query.\nError: %v", err)
+		t.Logf("Query: %+v", mismatched)
+		t.Logf("Matched: %+v", cbPluginsIds)
+	}
+	cb.Clear()
+}
+
+// Test the query service ids
+func TestQueryServicesIds(t *testing.T) {
```
meganokeefe: if you have a standard `setup()` helper that you're using for `QueryServices`, `QueryPlugins`, etc. you can literally copy what you do in `QueryTest` for `QueryIds`. only difference is that your expectedOutput would be `[]string`, vs `[]struct`. 

https://github.com/sigma-dev/sigma/blob/sites-api/database/couchbase_test.go#L273

tests/integration/database/couchbase/couchbase_test.go
```diff
@@ -0,0 +1,901 @@
...
+		t.Fatalf("QueryServices Test Failed.\nCouldn't insert.\nError: %v", err)
+	}
+
+	cbServices, err := cb.QueryServices(model.Service{})
+	if len(cbServices) != 1 {
+		t.Errorf("QueryServices Test Failed.\nUnexpected error on empty query.\nError: %v", err)
+	} else if !servicesAreEqual(testSvc, cbServices[0]) {
+		t.Error("QueryServices Test Failed.\nMutated object.")
+	}
+	cbServices, err = cb.QueryServices(testSvc)
```
meganokeefe: ideally, the Query tests should Query every attribute separately, to test the `makeQuery()` logic for each type.

stephenrlouie:
can we rename the container to be called sigma-db, or couchbase. [Here](https://github.com/sigma-dev/sigma/blob/db-unit-tests/database/cbdeploy.sh#L4)

tests/integration/database/couchbase/couchbase_test.go
stephenrlouie: Not sure if we want to address this here. but if we insert multiple items with the same name we are in trouble. Not tested here.

https://github.com/sigma-dev/sigma/issues/54

tests/integration/database/couchbase/couchbase_test.go
```diff
@@ -0,0 +1,939 @@
...
+	// Verify plugin deleted
+	_, err = cb.GetPlugin(cas)
+	if err != gocb.ErrKeyNotFound {
+		t.Errorf("DeletePlugin Test Failed.\nUnexpected get on deleted.\nError: %v", err)
+	}
+	cb.Clear()
+}
+
+// Test deleting a service
+func TestDeleteService(t *testing.T) {
```
stephenrlouie: table test?

tests/integration/database/couchbase/couchbase_test.go
```diff
@@ -0,0 +1,939 @@
...
+	err = cb.DeleteService("")
+	if err != db.ErrEmptyID {
+		t.Errorf("DeleteService Test Failed.\nUnexpected error on deleting empty string.\nError: %v", err)
+	}
+	cb.Clear()
+}
+
+// Test deleting a service operation
+func TestDeleteServiceOperation(t *testing.T) {
+	cas, err := cb.InsertServiceOperation(testSvcOp)
```
stephenrlouie: table test?

stephenrlouie:
seconded a lot of megans feedback.

- [ ] tabletests anywhere you do something more than once
- [ ] negative test cases for Update, currently only happy path

tests/integration/database/couchbase/couchbase_test.go
vgunning: I won't add in tests for this since the fix is in the works.
But it does need to be something tested once implemented.

tests/integration/database/couchbase/couchbase_test.go
```diff
@@ -0,0 +1,1606 @@
...
+}
+
+func TestMain(m *testing.M) {
+	rc := m.Run()
+	os.Exit(rc)
+}
+
+// Tests starting a new couchbase with various configs
+func TestNewCouchbase(t *testing.T) {
+	ip := os.Getenv("")
```
mikeynap: I think you need to provide an argument to Getenv -- this should always return "".

PR #61, opened 2017-05-09
renamed Updated to lastUpdated

stephenrlouie:
### Change log

Rename `service.Updated` to `service.LastUpdated`

Set that value in `Insert` as well as `Update`

### Notes

Was not setting on insert, also think LastUpdated is a little more intuitive.

Made DB responsible for setting `LatUpdate = time.Now()` since it was spreading to API/server and manager, and figured everyone would forget if we ever edit the service to update that field


adamschaub:
What's with the additional doc changes?

stephenrlouie:
all the other changes are because the api was calling it Updated. calling it LastUpdated changed all those `api/client` things

PR #64, opened 2017-05-10
Log improvements

meganokeefe:
### Changelog 

- [x] Updated existing logs to log at the right level. Ie. one-and-done startup logs are now `Info`, `Debugs` that are the result of errors are now `Warn`, etc. 
- [x] Removed some `Debug` logs to reduce the volume of logs that occurs with every ticker. Most significantly, removed the `Generated Query` Debug logs that are the source of most of the volume.


manager/manager.go
```diff
@@ -274,7 +272,14 @@ func (m *manager) statusOperation(op model.ServiceOperation) {
 		log.WithError(err).Errorf("Could not get progress of %s", op.Id)
 		return
 	} else {
```
mikeynap: Not actually sure if this is wise.

manager/manager.go
```diff
@@ -291,7 +290,10 @@ func (m *manager) startOperation(op model.ServiceOperation) {
 		}
 		return
 	}
-	log.WithField("Service", op.Service.Name).Infof("Started Operation %s (%s)", op.Id, op.Operation)
+
+	if op.Operation != model.ServiceOperationStatus {
+		log.WithField("Service", op.Service.Name).Infof("Started Operation '%s'", op.Operation)
```
adamschaub: `%q` prints quotes around the string, if you're interested

PR #65, opened 2017-05-10
Config to error on unknown types

stephenrlouie:
### Change log

Modified how config handles fields it cannot reflect on, aka: anything we don't support like slices, maps, or custom types.


config/config.go
```diff
@@ -186,11 +192,15 @@ func eachSubField(i interface{}, fn func(reflect.Value, string, string)) {
 			field := t.Field(i)
 			for i := 0; i < field.NumField(); i++ {
 				if field.Field(i).CanSet() {
-					fn(t, name, field.Type().Field(i).Name)
+					fmt.Printf("Name: %v Field: %v\n", name, field)
```
adamschaub: Remove

config/config.go
```diff
@@ -104,41 +105,44 @@ func getCfg(gCfg *GlobalConfig) {
...
 			case reflect.Float64:
 				v := viper.GetFloat64(str)
 				if v == 0 {
-					return
+					return nil
 				}
 				subField.SetFloat(v)
 				// TODO: Consider additional types (slice, map, etc)
+			default:
+				return fmt.Errorf("%s is unsupported by config", subField.Type().String())
```
adamschaub: Let's also list the config struct/field name here. 

```
--- Sigma encountered an Error: ---
	[]string is unsupported by config
```

Could be a little more explicit

config/config.go
stephenrlouie: good catch!

config/config.go
stephenrlouie: updated to include, FieldName and SubFieldName

api/v1/server/server.go
```diff
@@ -25,6 +25,7 @@ const UpdateRetryTimeout = 5 * time.Second
 
 type Config struct {
 	URL string
+	Ha  []string
```

PR #71, opened 2017-05-11
Service site defaults to local

meganokeefe:
Sigma no longer throws an error when a Service is added with an empty site. Defaults to Local.


model/model.go
```diff
@@ -17,9 +17,9 @@ import (
 	"github.com/pborman/uuid"
 )
 
-//The constant to search on for ID
 const (
-	ID = "Id"
+	ID    = "Id" //The constant to search on for ID
+	Local = "Local"
```
adamschaub: Probably prefer `DefaultSite`

PR #73, opened 2017-05-11
Integration tests run on substring `[integration]`

stephenrlouie:
### Change log

- Support for travis to run integration test when the substring `[integration]` is found in a commit message. (See my commit history for example)


adamschaub:
hot

PR #74, opened 2017-05-11
Changed froms to only use serviceName

mikeynap:


manager/manager.go
```diff
@@ -514,39 +514,23 @@ func (m *manager) collectInput(svc *model.Service, plugin model.Plugin) (bool, e
 			// Collect input from other services
 			for _, from := range item.From {
 				pts := strings.Split(from, ".")
-				if len(pts) < 2 {
-					log.WithError(ErrFromFormatInvalid).Errorf("Service %s Input %q invalid. Must have three parts", svc.Name, from)
+				if len(pts) < 1 {
```
adamschaub: Still need to check `len(pts)` is at least length 2, because of `svcName` and `varName` below

mikeynap: wait why was it 2 before lol.

manager/manager.go
```diff
@@ -514,39 +514,23 @@ func (m *manager) collectInput(svc *model.Service, plugin model.Plugin) (bool, e
...
+					if outputService, err = m.db.GetService(svcName); err != nil {
+						log.WithError(err).Warnf("Could not get service %s", svcName)
 						set = false
 						continue
-					} else {
-						outputService = services[0]
 					}
-				} else {
-					outputService = foundSvc
+					svcs[svcName] = outputService
```
adamschaub: unique service names makes this so much cleaner...

mikeynap: haha right

PR #79, opened 2017-05-12
Manager TODOs

adamschaub:
Fixes to manager to handle error cases.

TODO:
(list as they are found)
- [x] When operation fails to start, appropriately set state to failed
- [x] When no output/progress/status can be obtained from the operation, set state to failed and remove
- [ ] Make retry values configureable


manager/manager.go
```diff
@@ -204,53 +201,85 @@ func (m *manager) updateServices() {
...
+	}()
 
-		switch service.Status {
-		case model.ServiceStatusInactive:
-			// Run start
-			m.startService(&service, plugin)
+	plugin, err := m.pluginFor(service)
+	if err != nil {
+		log.WithError(err).Errorf("Could not get plugin for service %q", service.Name)
+		service.StatusDescription = "Provided plugin doesn't exist! YOU FOOL"
```
mikeynap: I do enjoy the cheekiness, but... 
If you wanna keep this I'm fine, just playing Megan's Advocate.

manager/manager.go
```diff
@@ -301,56 +330,73 @@ func (m *manager) updateOperations() {
 }
 
 func (m *manager) startOperation(op model.ServiceOperation) {
+	defer func() {
```
mikeynap: Maybe we should make this a function that gets deferred in each of these bodies? These are identical across the manager functions, could be a case for un-inlining.

model/bool.go
```diff
@@ -0,0 +1,9 @@
+package model
+
+func NewBool(b bool) *bool {
+	return &b
+}
+
+func BoolValue(b *bool) bool {
```
mikeynap: I like this, but would it make more sense to have an Enabled() function on the service? So `s.IsEnabled()` instead of `model.BoolValue(s.Enabled)`? I can see arguments for both, just figured I'd pose an alternative.

model/model.go
```diff
@@ -60,17 +62,49 @@ func (m *MetaImpl) Get(k string) (interface{}, bool) {
...
-		m.Metadata = make(map[string]interface{})
-	}
-	if v, ok := m.Metadata[k]; ok {
-		if v2, ok2 := v.(string); ok2 {
-			return v2, ok2
+	if v, ok := m.Get(k); ok {
+		switch v := v.(type) {
+		case int:
+			return fmt.Sprintf("%d", v), true
+		case int64:
```
mikeynap:
```	
               case int, int64:
			return fmt.Sprintf("%d", v), true
		case float32, float64:
			return fmt.Sprintf("%f", v), true
```

model/model.go
```diff
@@ -60,17 +62,49 @@ func (m *MetaImpl) Get(k string) (interface{}, bool) {
...
+			return "", false
 		}
 	}
 	return "", false
 }
 
+func (m *MetaImpl) GetInt(k string) (int, bool) {
+	if v, ok := m.Get(k); ok {
+		switch v := v.(type) {
+		case int:
```
mikeynap: sucks, but do we need int8,int16? Idk why we would be storing those... up to you.

model/bool.go
adamschaub: Can, I wanted these because we had other generic uses for BoolValue. I agree though

model/model.go
adamschaub: thx

model/model.go
adamschaub: Added those as well. Better safe than sorry.

manager/manager.go
adamschaub: Fixed

mikeynap:
the api/convert package does not currently set enabled when it converts to->from api objects.

PR #80, opened 2017-05-16
Bundle output

mikeynap:
This implements the ability to get the output of a "bundle" service. 
A service that starts a bundle doesn't actually contain any output values, but it can have output fields that reference it's internal (or external/other) services via the From: field.

For example, this service that starts a bundle (which contains one service called myBundlesService.OutputValue): 
```
{ 
   Name: "bundle_starting_service", BundleName: "myBundle", Inputs: [...], 
   Outputs: [ { 
          Name: OutputValue, 
          From: [myBundlesService.OutputValue, someOtherService.OutputValue]
     }]  
}
```

Querying `services/bundle_starting_service` now goes and fetches the values if they are available. If the From: specifies another bundleService, some recursion is needed until the actual end value is achieved.

The heavy lifting for this is done in `resolveFrom`, which has a bunch of tests for it (as well as tests in it's parent calling function and integration). 

It might be wise to read/review bundleV2 to get further context on this PR. Make sure this is merged BEFORE bundleV2.

Upon writing the docs, a few todos popped up:
- [x] NameSpacing services started from bundle -- since the names are defined in the bundle, multiple services for the same bundle would have conflicts.

Before reviewing, read up on bundles in the docs (gist: https://gist.github.com/mikeynap/9a0b880633950e69dad50a71958543a2)


config.yaml
```diff
@@ -3,7 +3,7 @@ Log:
   Level: debug
 Manager:
   OperationUpdateInterval: 10s
-  ServiceStatusInterval: 1m
+  ServiceStatusInterval: 2m
```
adamschaub: Smart, 1m is way too short

manager/manager.go
```diff
@@ -487,13 +502,13 @@ func (m *manager) createServiceOperation(svc *model.Service, plugin model.Plugin
...
 // CollectOutput. Updates Service outputs according to plugin and output
-func collectOutput(svc *model.Service, plugin model.Plugin, output map[string]interface{}) {
+func collectOutput(svc *model.Service, plugin model.Plugin, output map[string]interface{}) (changed bool) {
 	// Make sure plugin outputs are added to svc
 	out := model.CanonicalizeOutput(output)
 	for _, item := range plugin.Output {
 		exists := false
 		for _, output := range svc.Output {
-			exists = output.Name == item.Name || exists
+			exists = exists || output.Name == item.Name
```
adamschaub: NO

manager/manager.go
```diff
@@ -623,6 +658,93 @@ func (m *manager) collectInput(svc *model.Service, plugin model.Plugin) (bool, e
...
+func (m *manager) resolveFrom(svcName, varName string, trail map[string]bool) (model.Service, error) {
+	svc, err := m.db.GetService(svcName)
+	if err != nil {
+		return svc, err
+	}
+	if !svc.IsBundleService() { // If this is a normal service, nothing else to do.
+		return svc, nil
+	}
+
+	if trail == nil { //
```
adamschaub: NIT //

manager/manager.go
```diff
@@ -623,6 +658,93 @@ func (m *manager) collectInput(svc *model.Service, plugin model.Plugin) (bool, e
...
+			// keep recursing until either a normal service or a cycle is found.
+			fromSvc, err := m.resolveFrom(svcN, fromN, trail)
+			if err != nil {
+				return svc, err
+			}
+
+			// svcN_fromN gave a real value, so remove it from the trail
+			// If we don't do this, we cannot resolve the same k_v more
+			// than once.
+			trail[svcN+"_"+fromN] = false
```
adamschaub: Write a function for this, then write a test for that function

manager/manager.go
```diff
@@ -813,20 +960,33 @@ func (m *manager) populateBundleInput(bs model.Service, in *model.ServiceDataIte
...
 	}
 	return nil
 }
+
+func parseFrom(svc string) (svcname, varname string, err error) {
+	parts := strings.Split(svc, ".")
+	if len(parts) < 2 {
+		return "", "", ErrFromFormatInvalid
+	}
+	svcname, varname = strings.Join(parts[0:len(parts)-1], "."), parts[len(parts)-1]
```
adamschaub: Do we have this documented somewhere?

database/couchbase.go
```diff
@@ -81,18 +81,20 @@ func NewCouchbase(cfg CouchbaseConfig) (*Couchbase, error) {
...
+		h = "//" + h
+	}
+	if _, err = url.Parse(h); err != nil {
 		return nil, err
-	} else {
-		cb.cfg.host = *u
 	}
 
-	cb.DB, err = gocb.Connect(cb.cfg.host.String())
+	cb.DB, err = gocb.Connect(cfg.Host)
```
adamschaub: Do we need `//` for this connect, or just for parse?

manager/manager.go
mikeynap: that's my favorite comment.

manager/manager.go
mikeynap: Yes, it's actually live on the docs now -- must have gotten merged.

database/couchbase.go
mikeynap: Just for parse -- not for connect.
Idk how necessary having parse be at the top is anyway -- but the // is for parse.

PR #81, opened 2017-05-16
Messaging Queues

adamschaub:
# TODO
Lots:
- [x] Deploy/stop scripts for running message queue for tests
- [x] Make sure updates to services/operations are idempotent. Important for handling potential message duplication
- [x] Tie operation updates to service queue consumption
- [x] Update/add manager tests
- [x] Rabbit Impl
- [x] Leader election for svc init duty
- [x] Spin off goroutine per service, limit max goroutines

# Note
- The service update loop is locked to every 10 seconds. Not configureable right now, given that we don't have any concept of distributed configuration. Could be added if we are okay with some managers causing longer/shorter refresh rates.

# Future
- [ ] Identify points for backpressure


manager/manager.go
```diff
@@ -78,13 +80,17 @@ type manager struct {
 	tasker task.Tasker
 }
 
-func NewManager(db database.Database, tasker task.Tasker, cfg Config) (*manager, error) {
+func NewManager(db database.Database,
+	tasker task.Tasker,
```
meganokeefe: why are these params on different lines

config/config.go
```diff
@@ -41,6 +42,9 @@ func InitFlags(sigmaCmd *cobra.Command) error {
 	sigmaCmd.PersistentFlags().Bool("nulldb", false, "Use the in-memory nulldb for storage")
 	viper.BindPFlag("nulldb", sigmaCmd.PersistentFlags().Lookup("nulldb"))
 
+	sigmaCmd.PersistentFlags().Bool("channel-init", false, "Use the in-memory queue for service initiation")
+	viper.BindPFlag("nulldb", sigmaCmd.PersistentFlags().Lookup("channel-init"))
```
meganokeefe: you're binding channel-init to null-db here, FYI

manager/manager.go
```diff
@@ -136,10 +142,15 @@ func (m *manager) Start() error {
 			m.processBundles()
 
 			// Check for dirty/inactive services
-			m.updateServices()
-
-			// Follow up on outstanding ServiceOperations
-			m.updateOperations()
+			for {
```
meganokeefe: Clarification Q -- so on every ticker, a sigma manager will ingest the entirety of its service queue, and process all of them. And only exits the `for` loop once the queue is empty?

Is there any situation where it could get trapped in this loop?

manager/service_provider.go
```diff
@@ -0,0 +1,89 @@
...
+	return &managerServiceProvider{
+		db,
+		svcInit,
+		svcProvide,
+		make(chan struct{}),
+	}
+}
+
+// Start implements service to begin the manager
+func (m *managerServiceProvider) Start() error {
```
meganokeefe: Would it make sense to use steve's Services package here, or nah?

manager/service_provider.go
```diff
@@ -0,0 +1,89 @@
...
+	for {
+		select {
+		case <-m.quit:
+			log.Info("Quitting Sigma Service Provider")
+			return nil
+		default:
+			if m.svcInit.Wait(time.Millisecond * 20) {
```
meganokeefe: why have this `Wait()`? look at the func definition and not totally understanding its purpose. 

also, reasoning behind having it be 20ms?

manager/service_provider.go
```diff
@@ -0,0 +1,89 @@
...
+		if err != nil {
+			log.WithError(err).Errorf("Unable to queue services")
+			return
+		}
+		if n == 0 {
+			failToQueue++
+			log.Debug("Failed to queue any services. Retrying")
+			continue
+		}
+		services = services[n:]
```
meganokeefe: You're chopping off part of `services` here -- why?

queue/channels.go
```diff
@@ -0,0 +1,97 @@
...
+		ch: make(chan struct{}),
+	}
+	go func() {
+		for {
+			// Attempt to send every interval, forever
+			select {
+			case <-time.After(interval):
+				// Send on the channel, or else
+				select {
+				case ci.ch <- struct{}{}:
```
meganokeefe: why use a `select` here if you don't act on either of the cases?

manager/service_provider.go
adamschaub: `Send` may not send all of the services at once. It returns a number which indicates that the first `n` services were queued successfully. So I trim them from the beginning, and try to send the remaining until none remain.

queue/channels.go
adamschaub: It tries to send a message on `ci.ch`. If it can't do that immediately (a message has already been sent, and not yet received) it `default`s and does nothing.

manager/service_provider.go
adamschaub: 20ms was just an arbitrary decision.  `Wait` returns true if this service received a signal to queue services. If it doesn't receive anything in 20ms, it passes and tries again.

manager/service_provider.go
adamschaub: I am using Steve's package

manager/manager.go
adamschaub: It would only get "trapped" if it had a constant incoming stream of services to update that it couldn't keep up with. As soon as there are no services to update, the loop will quit. 

Per Matt's guidance, this is going to be changed out for a worker pool pattern, where we have a max of N goroutines that can update services. So this will change a bit soon.

config/config.go
adamschaub: good catch

manager/manager.go
adamschaub: Trying out a different format. I'm thinking this helps readability when the inputs get rather long.

mikeynap: we C89 now bois.

manager/service_provider.go
mikeynap: ;)

cmd/sigma.go
```diff
@@ -66,20 +67,45 @@ func sigmaRun(cmd *cobra.Command, args []string) {
...
+
+	// Queues
+	var svcProvide queue.ServiceProvider
+	var svcConsume queue.ServiceConsumer
+	if gCfg.UseChannelInit {
+		channelQueue := queue.NewChannelQueue(10)
+		svcConsume = channelQueue
+		svcProvide = channelQueue
+	} else {
+		if provide, err := queue.NewRabbitServiceProvider(gCfg.Rabbit); err != nil {
```
mikeynap: err already defined at top;
```	var svcProvide queue.ServiceProvider
	var svcConsume queue.ServiceConsumer
	if gCfg.UseChannelInit {
		channelQueue := queue.NewChannelQueue(10)
		svcConsume = channelQueue
		svcProvide = channelQueue
	} else {
		if svcProvide, err = queue.NewRabbitServiceProvider(gCfg.Rabbit); err != nil {
			globalErr = fmt.Errorf("Couldn't create rabbit provider: %v", err)
			return
		}		
		if svcConsume, err = queue.NewRabbitServiceConsumer(gCfg.Rabbit); err != nil {
			globalErr = fmt.Errorf("Couldn't create rabbit consumer: %v", err)
			return
		}	
	}
```

mikeynap: idk why I feel a strong need to be pedantic and minimalistic in main() ..

adamschaub: I like it

config/config.go
```diff
@@ -15,16 +15,19 @@ import (
...
-	Manager   manager.Config
-	UseNulldb bool
-	Server    server.Config
+	Log            log.Config
+	Docker         task.DockerConfig
+	Couchbase      database.CouchbaseConfig
+	Manager        manager.Config
+	Rabbit         queue.RabbitConfig
+	UseNulldb      bool
+	UseChannelInit bool
```
mikeynap: Can we have a less ugly name

config/config.go
```diff
@@ -41,6 +44,9 @@ func InitFlags(sigmaCmd *cobra.Command) error {
 	sigmaCmd.PersistentFlags().Bool("nulldb", false, "Use the in-memory nulldb for storage")
 	viper.BindPFlag("nulldb", sigmaCmd.PersistentFlags().Lookup("nulldb"))
 
+	sigmaCmd.PersistentFlags().Bool("channel-init", false, "Use the in-memory queue for service initiation")
+	viper.BindPFlag("channel-init", sigmaCmd.PersistentFlags().Lookup("channel-init"))
```
mikeynap: should this be true or false to start? Should we add logic to "fall back" on this if rabbit config is not specified?

adamschaub: I'd rather not, because it's mostly just for testing/standalone. And you can foreseeably get along with the rabbit config defaults.

manager/service_provider.go
```diff
@@ -0,0 +1,95 @@
...
+func (m *managerServiceProvider) Stop() {
+	close(m.quit)
+}
+
+func (m *managerServiceProvider) queueServices() {
+	// TODO: Naive for now. Can get fancier maybe? Ignore services that
+	// have been recently updated? Ignore non-enabled services? Add services with outstanding operations?
+	var services []model.Service
+	var err error
+	if services, err = m.db.QueryServices(model.Service{}); err != nil {
```
mikeynap: it would be wayyyyy nicer if we did a QueryServiceIDs here. No where along the path from here-->manager.updateService() do we need anything other than the ID.

Getting _all_ the services and then sending them through the queue, then unmarshalling them is a lot of work -- especially since we just do a get immediately anyway on the other end. 

This would require changing manager.updateService() to take a string instead of a Service.

PR #82, opened 2017-05-16
NomadTasker 

meganokeefe:
NomadTasker implements the Tasker interface, and is responsible for translating `ServiceOperations` into Nomad Jobs. Each Nomad Job has one TaskGroup with two tasks: `serviceTask` is the service deployment, and `shipperTask` gathers and ships logs/progress/outputs back to the sigma API. 

## Changelog 

- [x] vendored the Nomad API 
- [x] nomad deploy + stop scripts, added to the integration test script 
- [x] added a modified version of Victoria's psi vagrant/nomad environment (they use ubuntu trusty now, not Centos)
- [x] implemented the nomad Tasker in `task/nomad.go`.
- [x] `Services` now have a `savedir` 
- [x] model defs and DB ops for `Shipper` 
- [x] added API funcs for `AddServiceShipper` and `GetServiceShipper` + regenerated client 
- [x] updated the helloworld `main.go` plugin to use 2 environment variables, for the 2 savedirs (see Notes below) 
- [x] `plugin.go` functions all take in a `savedir` parameter 
- [x] nomad config 
- [x] Couchbase tests for Shipper 
- [x] Convert tests for `Shipper` 
- [x] API tests for AddServiceShipper and GetServiceShipper 
- [x] basic nomadtasker integration tests 
- [x]  configuration for travis to locally install and run Nomad for integration tests (can't use a container due to the volumes problem)
 
## Notes 

Due to Nomad's resource isolation between Tasks and TaskGroups, and the way Nomad dynamically creates an allocation directory, it seems impossible to have inputs, outputs, and progress all in one place. Up until now, we've been using one hardcoded savedir, `/var/lib/sigma`, in our plugins.

But it seems that Nomad will need 2 savedirs. Inputs can be mounted from anywhere, but `progress` and `outputs` must get written to `alloc/` in order for the ServiceShipper to access them. (ServiceShipper is a separate task, and `alloc/` is the only shared directory between tasks). 

After discussing with @adamschaub, deciding to give all Services a `savedir` parameter, which will be passed around to the plugin as two environment variables-- one for input, one for output. For non-Nomad taskers, `savedir` will refer to where inputs, logs, progress, and output live, and will default to `/var/lib/sigma`. So functionally, for non-Nomad taskers, `savedir` refers to both inputSavedir and outputSavedir. 

But for NomadTasker, `savedir` routes to inputSavedir and we hardcode outputSavedir to be `/alloc.`

## TODO / for separate PRs 
1. nomad client mock to use for unit tests -- similar to Adam's `mock_transport` https://github.com/sigma-dev/sigma/blob/master/task/mock_transport.go 
2. nomad tasker unit tests 
3. update the sigmadev/hello plugin on Dockerhub to use two savedir env variables, plus use the actual sigma `plugin/client` instead of the random one I have on my personal github
4. update the plugin developer guide to note the existence of the two savedirs. plus update the docs to note the existence of a new tasker  
5. Update dockerTasker to pass the savedir environment variable(s) to the plugin, and to have a configurable client URL 
6. Overhaul the way the Tasker and Shipper works, per 5/17 discussion with @mcaulfield and @adamschaub. Will break up userAPI service operation endpoint such that logs can be queryable by serviceOperation ID. And progress and Outputs will live in the Service DB struct. Those functions will be removed from Tasker-- manager queries DB for these. In terms of NomadTasker specifically, this change will mean that the "shipper" task will curl twice-- once to update the Logs, and once to update the Progress/Outputs.


mikeynap:
in a clean env, this does not build. It needs `github.com/mitchellh/go-homedir`.

examples/hello-world/main.go
```diff
@@ -20,9 +25,26 @@ type HelloWorldConf struct {
 	Name string `json:"Name"`
 }
 
+// pull in outputSavedir environment variables
```
mikeynap: Can we add this to plugin/client? That is, parsing input/output savedir from environment? 
Also, it should really be SIGMA_INPUT_SAVEDIR or something.

examples/hello-world/main.go
```diff
@@ -117,13 +139,16 @@ func fail(err string, progress string, exitCode int) {
 		fmt.Fprintln(os.Stderr, err)
 	}
 	if progress != "" {
-		plugin.WriteProgress(progress)
+		plugin.WriteProgress(progress, outputSavedir)
```
mikeynap: Maybe forgo the argument here, and have WriteProgress look for it in the environment, else use default. 

This logic should be baked into the client SOMEWHERE to make it 'official'.

task/nomad.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// Status gets the nomad status for this Service + translates to a Sigma status
+// Note that Status "stopped" is unused here. There isn't really a Nomad equivalent.
+func (nt *NomadTasker) Status(so *model.ServiceOperation) (string, error) {
+	inf, _, err := nt.Client.Jobs().Info(so.Service.Name, nil)
+	if err != nil {
+		return "", err
+	}
+	stat := *inf.Status
```
mikeynap: can inf.Status be nil?

task/nomad.go
```diff
@@ -0,0 +1,236 @@
...
+
+	// Service Env
+	env := map[string]string{}
+	if _, ok := so.Plugin.Config["env"]; ok {
+		if envMap, ok2 := so.Plugin.Config["env"].(map[string]string); ok2 {
+			env = envMap
+		}
+	}
+	//Add two Savedirs to task env. we're hardocoding the output savedir so that the shipper can access the files
+	env["INPUT_SAVEDIR"] = so.Service.SaveDir
```
mikeynap: SIGMA_

task/nomad.go
```diff
@@ -0,0 +1,236 @@
...
+	tg.AddTask(serviceTask)
+	tg.AddTask(shipperTask)
+	j.AddTaskGroup(tg)
+
+	return j, nil
+}
+
+// populateShipper creates the sidecar script that curls logs/progress/outputs for job <id> and <name> to the sigma API <endpoint>
+func populateShipper(id string, name string, endpoint string) string {
+	return fmt.Sprintf(`openBracket={;
```
mikeynap: beautiful.

task/nomad.go
```diff
@@ -0,0 +1,236 @@
...
+	noDashes=$(echo $noReturn| tr ' ' '-');
+	prog=\"$noDashes\";
+	rawOut=$(cat alloc/output);
+	baseOut=$(echo -n $rawOut | base64);
+	noNewlines=$(echo $baseOut| tr '\n' '-');
+	noReturn=$(echo $noNewlines| tr '\r' '-');
+	noDashes=$(echo $noReturn| tr ' ' '-');
+	out=\"$noDashes\";
+	closeBracket=};
+	jsonContents=$openBracket$idLabel$id$nameLabel$name$logsLabel$logs$progLabel$prog$outLabel$out$closeBracket;
```
mikeynap: sexy.

task/nomad.go
```diff
@@ -0,0 +1,236 @@
...
+	echo $jsonContents;
+	url=%v/v1/services/%v/shipper;
+	echo $url;
+	curl -X POST --data $jsonContents $url;
```
mikeynap: hot.

examples/hello-world/main.go
meganokeefe: Agreed!

stephenrlouie:
@meganokeefe have we recorded those TODOs somewhere?

examples/hello-world/main.go
```diff
@@ -7,30 +7,30 @@ import (
 	"strings"
 	"time"
 
-	plugin "github.com/sigma-dev/sigma/plugin/client"
+	client "github.com/meganokeefe/pluginsavedir"
```
stephenrlouie: why not submit these changes to `sigma-dev/sigma/client`?

examples/hello-world/main.go
```diff
@@ -117,19 +117,22 @@ func fail(err string, progress string, exitCode int) {
...
 	if progress != "" {
-		plugin.WriteProgress(progress)
+		client.WriteProgress(progress)
 	}
 	os.Exit(exitCode)
 }
 
 func main() {
-	inputs, err := plugin.GetInputs()
+	time.Sleep(time.Second * 10)
```
stephenrlouie: side note for another time, I want to make this sleep time configurable. would make testing a lot nicer.

plugin/client/plugin.go
```diff
@@ -19,6 +17,27 @@ const (
 	PluginArgUndefined           = "undefined"
 )
 
+var (
+	inputSavedir  string
+	outputSavedir string
+)
+
+// parse savedir environment variables + init defaults if unset
+func init() {
```
stephenrlouie: could we use this? or does it still differ from https://github.com/meganokeefe/pluginsavedir/blob/master/plugin.go

task/nomaddeploy.sh
```diff
@@ -0,0 +1,87 @@
+if [ -z $1 ]; then
+  echo "NOT JENKINS"
```
stephenrlouie: NIT, remove JENKINS. One Exposes ports and the other does not, not certain if we need it now.

Also whats the difference between the first two, besides stop and start? You could just call a nomad-stop.sh here in case that gets more complex in the future.

tests/integration/setup.sh
```diff
@@ -1,2 +1,20 @@
 echo "Integration tests dependency set up"
-./database/cbdeploy.sh
\ No newline at end of file
+./database/cbdeploy.sh
+#./task/nomaddeploy.sh
```
stephenrlouie: rm?

tests/integration/shutdown.sh
```diff
@@ -1,2 +1,9 @@
 echo "Integration tests dependency shutdown"
-./database/cbstop.sh
\ No newline at end of file
+
+# Kill nomad agent running on the travis host
+echo "Killing nomad..."
+ps -ef | grep nomad | grep -v "grep" | awk '{print $2}' | sudo xargs kill;
```
stephenrlouie: `[pgrep](https://linux.die.net/man/1/pgrep) nomad`?

tests/integration/shutdown.sh
```diff
@@ -1,2 +1,9 @@
 echo "Integration tests dependency shutdown"
```
stephenrlouie: I think we should clean up the nomad stuff too since we add it on `setup.sh`

tests/integration/task/nomad/nomad_test.go
```diff
@@ -0,0 +1,352 @@
+// Integration tests for the NomadTasker.
+// TODO - remove the couchbase dependency from these tests, once we overhaul the way Tasker works
```
stephenrlouie: we should get this into rally or issues?

config-nomad.yaml
```diff
@@ -0,0 +1,21 @@
+---
+Log:
+  Level: debug
+Manager:
+  OperationUpdateInterval: 30s
+  ServiceStatusInterval: 1m
+  ServiceStatusInterval: 30s
```
stephenrlouie: duplicate?

config-nomad.yaml
```diff
@@ -0,0 +1,21 @@
...
+Nomad:
+  ClientUrl: 127.0.0.1:4646
+  APIUrl: 127.0.0.1:24601
+Couchbase:
+  Host: 10.131.32.127
```
stephenrlouie: localhost please

plugin/client/plugin.go
mikeynap: it's the same.

mikeynap: changed to use sigma version

PR #87, opened 2017-05-18
Https user api

vgunning:
### PR checklist

- [X] Read the [contribution guidelines](https://www.google.com). TODO: fill this in with proper link
- [X] Filed the PR against the correct branch: master

### Change log
User API Server config now contains:
- [x] Host: ip address
- [x] Port: http port
- [x] SSLPort: https port
- [x] Cert: file path to cert file
- [x] Key: file path to key file
The server stands up either http/https or both depending on what is provided in config. The server checks for existing cert/key if a sslport is provided.
- [x] Update the sigma config files
- [x] Update mustache files to auto-generate the additional client code.
- [x] Provide example https config with certs/key.

User API Client: has a `Configuration.AddCertFile(string) error` to add certs to an existing api. The basepath can be changed or scheme to https to generate the correct requests.

### Notes

If we want to add to the https system, we can setup separate router functions. Currently uses the same routes and process as the http routes.
The function sets up a separate router and serve so we can cleaning exist and shutdown the listener.
Go has a script to generate self-signed certs with OpenSSL in the tls library.

### TODO

- [ ] Possibly update mustache files for server.


api/v1/cert.pem
```diff
@@ -0,0 +1,18 @@
+-----BEGIN CERTIFICATE-----
```
meganokeefe: should this cert have been committed?

api/v1/api_command.go
```diff
@@ -0,0 +1,18 @@
+package main
```
meganokeefe: is this file a test? in which case, can it be moved to a separate unit test file?

api/v1/server/server.go
```diff
@@ -56,67 +68,162 @@ func MsgFromErr(typ string, err error) string {
...
 	}
-	s := &Server{cfg.URL, newRouter(), db, nil, make(chan struct{})}
-	s.router.InitRoutes(s)
 	return s, nil
 }
 
 func (s *Server) listen() error {
 	// Setting up listener for webhooks
-	listener, err := net.Listen("tcp", s.url)
+	listener, err := net.Listen("tcp", s.host+":"+s.port)
```
meganokeefe: nit, use `fmt.Sprintf` to assemble the URL

api/v1/api_command.go
vgunning: It was an example file I was using to make sure it worked.
Tests should be created. And this file should be removed.

api/v1/server/server.go
```diff
@@ -56,67 +68,162 @@ func MsgFromErr(typ string, err error) string {
...
+	}
+	s.secureListener = listener
+	return nil
+}
+
 //Start implements the service interface and starts the HTTP server
 func (s *Server) Start() error {
-	if err := s.listen(); err != nil {
-		return err
+	if s.port != "" {
```
meganokeefe: Should `Start()` return errors if one of `port` and/or `sslPort` are empty? This way you wouldn't have to wrap your `go func`s in if-statements?

api/v1/gen/templates/client/api.mustache
```diff
@@ -93,7 +93,7 @@ func (a {{{classname}}}) {{{nickname}}}({{#allParams}}{{paramName}} {{{dataType}
 	{{/hasQueryParams}}
 
 	// to determine the Content-Type header
-	localVarHttpContentTypes := []string{ {{#consumes}}"{{{mediaType}}}", {{/consumes}} }
+	localVarHttpContentTypes := []string{ {{#consumes}}"{{{mediaType}}}" {{/consumes}} }
```
mikeynap: You can't do this -- len(contentTypes) might be > 1, in which case the commas are needed.

api/v1/server/server.go
```diff
@@ -56,67 +68,165 @@ func MsgFromErr(typ string, err error) string {
...
-		return nil, fmt.Errorf("Invalid URL: %s", cfg.URL)
+	var err error
+	url := new(url.URL)
+
+	if url, err = url.Parse(cfg.Host); err != nil {
+		log.Errorf("Invalid Host: %v", cfg.Host)
+		return nil, ErrInvalidHost
+	}
+	s := &Server{
+		router:         newRouter(),
```
mikeynap: NIT: call newRouter() when needed -- like `if cfg.Port != "" { s.router = newRouter(); s.router.initRoutes(s)}`

mikeynap: Request:
for some reason I wrote newRouter() (line router.go:27) as 	return &Router{mux.NewRouter().StrictSlash(true)}. 

Could you change this to StrictSlash(false)? 

I'm not sure why we have it to true.

api/v1/server/server.go
```diff
@@ -56,67 +68,165 @@ func MsgFromErr(typ string, err error) string {
...
+		Port: "24601",
+	}
 }
 
 //New creates a new http server instance by taking in a url to listen on and a database object
 func New(cfg Config, db database.Database) (*Server, error) {
-	if _, err := url.Parse(cfg.URL); err != nil {
-		return nil, fmt.Errorf("Invalid URL: %s", cfg.URL)
+	var err error
+	url := new(url.URL)
```
mikeynap: no need for new, var url url.URL

api/v1/server/server.go
```diff
@@ -56,67 +68,165 @@ func MsgFromErr(typ string, err error) string {
...
+		secureRouter:   newRouter(),
+		port:           cfg.Port,
+		db:             db,
+		listener:       nil,
+		secureListener: nil,
+		tlsConfig:      tls.Config{},
+		quit:           make(chan struct{}),
+	}
+	var port string
+	s.host, port, err = net.SplitHostPort(url.Host)
```
mikeynap: Why are you requiring both cfg.Host to have a port AND cfg.port to be set? 
Like we parse the port of url.Host, but then never use it.

api/v1/server/server.go
```diff
@@ -56,67 +68,165 @@ func MsgFromErr(typ string, err error) string {
...
+	if port != "" {
+		return nil, ErrSeparatePort
+	}
+	if (cfg.Port == "") && (cfg.SSLPort == "") {
+		return nil, ErrNeedHTTPOrHTTPS
+	}
+
+	// Init the HTTP router if needed
+	if cfg.Port != "" {
+		s.router.InitRoutes(s)
```
mikeynap: All of this splitting functionality between Router and SecureRouter and listener/secureLIstener is unneeded and repetitive. 

Just test if it's HTTP or HTTPS, then setup router/listener with the correct struct.

api/v1/server/server.go
```diff
@@ -56,67 +68,165 @@ func MsgFromErr(typ string, err error) string {
...
+	}
+	s.secureListener = listener
+	return nil
+}
+
 //Start implements the service interface and starts the HTTP server
 func (s *Server) Start() error {
-	if err := s.listen(); err != nil {
-		return err
+	if s.port != "" {
```
mikeynap: Combine the listen functions (maybe have a secure bool argument) so you can get rid of all this CRUD.

api/v1/server/server.go
```diff
@@ -56,67 +68,165 @@ func MsgFromErr(typ string, err error) string {
...
 	var serveWg sync.WaitGroup
-	go func() {
-		serveWg.Add(1)
-		log.Infof("API Server listening on %s", s.url)
-		log.Debugf("TCP Listener: %v", http.Serve(s.listener, s.router))
-		serveWg.Done()
-	}()
+	if s.port != "" {
+		go func() {
+			serveWg.Add(1)
```
mikeynap: same.

api/v1/server/server.go
```diff
@@ -56,67 +68,165 @@ func MsgFromErr(typ string, err error) string {
...
+			log.Infof("API HTTP Server listening on %s", fmt.Sprintf("%s:%s", s.host, s.port))
+			log.Debugf("TCP Listener: %v", http.Serve(s.listener, s.router))
+			serveWg.Done()
+		}()
+	}
+	if s.sslPort != "" {
+		go func() {
```
mikeynap: ^^^

api/v1/server/server.go
vgunning: The logic is making sure that there isn't a port on the host url. It should either be specified as Port (HTTP) or SSLPort (HTTPS).

api/v1/server/server.go
vgunning: As far as I know, we're supporting being able to do HTTP, HTTPS, or both.

api/v1/server/server.go
```diff
@@ -56,41 +66,103 @@ func MsgFromErr(typ string, err error) string {
...
+	return Config{
+		Host: "127.0.0.1:24601",
+	}
 }
 
 //New creates a new http server instance by taking in a url to listen on and a database object
 func New(cfg Config, db database.Database) (*Server, error) {
-	if _, err := url.Parse(cfg.URL); err != nil {
-		return nil, fmt.Errorf("Invalid URL: %s", cfg.URL)
+	if (!strings.HasPrefix(cfg.Host, "http://")) && (!strings.HasPrefix(cfg.Host, "https://")) {
```
mikeynap: We got rid of this check -- remove lines 98-112 completely

api/v1/server/server.go
```diff
@@ -56,41 +66,103 @@ func MsgFromErr(typ string, err error) string {
...
+
 	return s, nil
 }
 
 func (s *Server) listen() error {
 	// Setting up listener for webhooks
-	listener, err := net.Listen("tcp", s.url)
-	if err != nil {
-		return err
+	if !s.secure {
```
mikeynap: Use this instead:
```
	var err error
	if !s.secure {
		s.listener, err = net.Listen("tcp", fmt.Sprintf("%s:%s", s.host, s.port))
	} else {
		s.listener, err = tls.Listen("tcp", fmt.Sprintf("%s:%s", s.host, s.port), &s.tlsConfig)
	}
	if err != nil {
		return err
	}
	return nil
```

api/v1/server/server.go
```diff
@@ -101,16 +173,26 @@ func (s *Server) Start() error {
 	}
```
mikeynap: We know that if s.listen() == nil that s.listener is not nil. You can remove all the nil checks below.

mikeynap:
If you'd merge in master I'll give her the ole merge.

PR #92, opened 2017-05-23
Properly remove service when API delete is issued

adamschaub:
On an service DELETE, set a `removed` flag. Cause manager to delete the service (after any existing operation has completed first)


manager/manager.go
```diff
@@ -247,6 +247,12 @@ func (m *manager) updateService(service model.Service) {
 		inputsChanged = changed
 	}
 
+	if model.BoolValue(service.Remove) {
```
stephenrlouie: why not just add another Status?

Issue #86, opened 2017-05-17
Remove the distinction between Physical and Logical Sites


meganokeefe:
Generalize the Site definition. No Physical vs. Logical distinction, all sites can become parents.

PR #99, opened 2017-05-26
Sites revamp 

### LINKED ISSUE #85: Sites Metadata
Sites should be able to have metadata to specify things like GPS coordinates.

meganokeefe:
Closes #85 and #86. 

## Changelog 

- [x] Removed the distinction between Physical and Logical sites. Now all sites can be parents. 
- [x] Sites now have `Attributes map[string]interface{}`. `Attributes` can store tasker-specific hints like the nomadNodeID, or geo coordinates of the edge site, etc. 
- [x] API Server's `UpdateSite` now prevents users from changing the `Site.Id`, which sigma autogenerates.
- [x] Service Operations now have a list of `model.Site` instead of a `Site` string. Manager is responsible for taking a Service's `Site` string and getting all the `Sites` that correspond to itself and its children. 
- [x] Updated NomadTasker to handle the new `ServiceOperation.Sites` list, to determine where to run the Job on. Currently, just chooses a `nomadNodeID` at random in the Sites list. Errors out if no nomad node ID is found.


api/v1/client/configuration.go
```diff
@@ -12,8 +12,11 @@ package apiclient
 
 import (
 	"crypto/tls"
+	//"crypto/x509"
```
adamschaub: remove

api/v1/client/docs/Site.md
```diff
@@ -5,8 +5,8 @@ Name | Type | Description | Notes
 ------------ | ------------- | ------------- | -------------
 **Id** | **string** |  | [optional] [default to null]
 **Name** | **string** |  | [optional] [default to null]
-**Physical** | **bool** |  | [optional] [default to null]
 **Parent** | **string** |  | [optional] [default to null]
+**Attributes** | [**map[string]interface{}**](interface{}.md) |  | [optional] [default to null]
```
adamschaub: What is `interface{}.md`

api/v1/client/service.go
```diff
@@ -1,28 +1,29 @@
...
 
-	Bundle string `json:"bundle,omitempty"`
-
 	Plugin string `json:"plugin,omitempty"`
 
 	PluginVersion string `json:"plugin_version,omitempty"`
 
 	Savedir string `json:"savedir,omitempty"`
 
+	Bundle string `json:"bundle,omitempty"`
```
adamschaub: k

api/v1/client/bundle.go
```diff
@@ -1,21 +1,24 @@
...
+
 	Id string `json:"id,omitempty"`
 
 	Name string `json:"name,omitempty"`
 
 	Services []Service `json:"services,omitempty"`
 
 	Input []PluginIo `json:"input,omitempty"`
+
+	Output []PluginIo `json:"output,omitempty"`
```
adamschaub: Weird, how wasn't this in here before?

api/v1/server/server.go
```diff
@@ -822,10 +822,13 @@ func (s *Server) UpdateSite(w http.ResponseWriter, r *http.Request) {
 		updatedSite := convert.FromAPISite(apiObj)
 		updatedSite.MetaImpl = site2Update.MetaImpl
 		updatedSite.Set("Dirty", "true")
-		updatedSite.Id = site2Update.Id
+		if updatedSite.Id != site2Update.Id {
+			log.Errorf(ErrBadSiteID)
+			return http.StatusBadRequest, fmt.Errorf(ErrBadSiteID)
```
adamschaub: Note: Addresses issue #107.

Also might think of making this a 409 "Conflict" response code.

manager/manager.go
```diff
@@ -823,20 +823,10 @@ func (m *manager) inputSatisfied(svc *model.Service, plugin model.Plugin) (satis
 
 // newOperation attempts to create a new ServiceOperation with the provided service.
 func (m *manager) newOperation(svc *model.Service, plugin model.Plugin, opType string) error {
-	// Process the service site. if Physical, use it. if Logical, pick a child.
-	chosenOne := svc.Site
-	givenSite, err := m.db.GetSite(svc.Site)
+	// Process the service's Site - gather all relevant sites (itself and its children, grandchildren, etc.)
+	childSites, err := m.gatherChildSites(svc.Site)
```
adamschaub: Me gusta. This really simplified this process.

manager/manager.go
```diff
@@ -870,46 +860,34 @@ func (m *manager) newOperation(svc *model.Service, plugin model.Plugin, opType s
...
-		//if physical, base case -> add me
-		if model.BoolValue(child.Physical) == true {
-			allPhysicalSites = append(allPhysicalSites, child.Name)
-		} else {
-			// if logical, get all of its physical sites
-			recursiveChildren, err := m.getAllPhysicalSites(child.Name)
+		// add me
+		allChildSites = append(allChildSites, child)
+		// get my children.
+		childChildren, _ := m.db.QuerySites(model.Site{Parent: child.Name})
```
adamschaub: Why this extra step? Can't you just `allChildSites = append(allChildSites, m.gatherChildSites(child.Name)...)`?

I'm not getting the need for the additional query/result check.

database/mock_database/mock_database.go
```diff
@@ -1,5 +1,5 @@
 // Automatically generated by MockGen. DO NOT EDIT!
-// Source: github.com/sigma-dev/sigma/database (interfaces: Database)
```
adamschaub: Why has this been updated?

api/v1/client/bundle.go
meganokeefe: No idea why.

api/v1/server/server.go
meganokeefe: Good point. 

And about that issue, yeah I debated shoehorning this duplicate checking for just the one struct, vs. all of them. But I think this check is probably in the right place? Steve has said to put as little logic in `database/` as possible.

database/mock_database/mock_database.go
meganokeefe: Because I changed the `model/` definition for `Site`, which propagates to the database + query functions.

api/v1/client/docs/Site.md
meganokeefe: Culprit:

```
        attributes:
          type: object
          additionalProperties:
            type: object
```

It appears in one other spot in the yaml (`plugin.config`) so I removed that too.

manager/manager.go
meganokeefe: You are correct, I was relying too much on the old template of this function / thinking for some reason that because I didn't have a recursive helper here, I had to do two queries.

Twas mistaken! updated.

database/mock_database/mock_database.go
adamschaub: Model definition shouldn't affect the mock. The only change to this file is reordering the function declarations.

PR #100, opened 2017-05-26
Docs updates for new Site definition

meganokeefe:
## Changelog

- [x] Removed Physical/Logical site documentation 
- [x] Added info about `Site.Attributes`
- [x] Added new API spec 
- [x] add information about site parentage and how we handle site hierarchies


_docs/api-spec.md
```diff
@@ -720,17 +680,22 @@ font-style: italic;
   "name" : "aeiou",
   "id" : "aeiou",
   "services" : [ {
+    "savedir" : "aeiou",
```
adamschaub: What's the deal with having savedir in the service? I'm assuming it's something added in the nomad update?

adamschaub:
LGTM

_docs/api-spec.md
meganokeefe: Right now, all Services now get a `Savedir`, which defaults to `/var/lib/sigma`. https://github.com/sigma-dev/sigma/blob/master/model/model.go#L189

https://github.com/sigma-dev/sigma/blob/master/task/task.go#L44

Underlying reason was that we probably shouldn't be hardcoding plugin/client to use `/var/lib/sigma`. So made it be an env variable. Technically two savedirs, input and output. https://github.com/sigma-dev/sigma/blob/master/plugin/client/plugin.go#L27

But: Nomad has to hardcode output savedir for reasons. So right now `Service.Savedir` corresponds to Nomad's InputSavedir only. 
And dockertasker is just using the default savedir right now and not configuring anything.

adamschaub: Alright. Doesn't affect this PR, but I'm not a huge fan, given that it is a detail that we ideally hide gritty details from the person making the service. SaveDir has to do with the Tasker/maybe the Plugin, not the service itself. I think the goal would be to have any Service run anywhere (any Tasker). This wouldn't be portable if we had another tasker type that was picky about mount locations.

Issue #98, opened 2017-05-25
Revamp plugin type handling


adamschaub:
Master issue for implementing more complex types per plugin.

What need do:
- [x] Add types to Plugin in model
- [x] Add types to Plugin in API
- [x] Update Plugin/Service inputs/outputs to be map of var-name -> var-info
- [x] Convert values to generic `interface{}`
- [x] Allow nested object/list types (i.e. list of objects that have lists of objects, etc)
- [ ] Add constraint processing for types
 - [x] Reflection on provided Plugin/Service values to ensure specified type matches actual value
- [x] Basic conversion to/from basic types. I.e. if a field specifies string and is given an int, convert the int to a string
- [x] Syntax to use specific vars for a nested object field
- [x] Add generic `validation` or `accept` to types (and inputs?)
- [x] Inputs optional

Separate into a few PRs:
- [x] Update model/api definitions (use maps instead of lists, use interface for values). Everything works the same otherwise. No strong vetting of types, etc
- [x] Type conversion functions (between all basic types + lists) + vet types/attempt conversion during processing
- [ ] Validation using contraints (add to processing). Update model/api accordingly. Just enum to start

mcaulfield:
Move on with your life.

PR #103, opened 2017-05-26
Model/API using generic types

adamschaub:
Ref: #98 

Update model/api definitions (use maps instead of lists, use interface for values). Everything works the same otherwise. No strong vetting of types, etc. 

- [x] Update types to generic interface{}
- [x] Provide initial support for merging interface values (just []string for now)
- [x] Convert all example plugin/service definitions to map

# Important Changes
Input/Outputs in Services/Plugins are now maps, NOT lists.
This:
```
{
  "input": [{
      "name": "Name",
      "type": "string",
      "value": "Unicorn"
    }
  ]
}
```
Now looks like this:
```
  "input": {
    "Name": {
      "type": "string",
      "value": "Unicorn"
    }
  }
}
```

Additionally, values are no longer constrained to strings/[]string. Reflection is used during gathering inputs/outputs to merge values together. See a useless example under `examples/hello-world/chain-service.json`


api/v1/convert/convert_test.go
```diff
@@ -214,18 +208,27 @@ func TestService(t *testing.T) {
...
 	}
 
 	statRet := ToServiceStatus(services.m)
-	if !reflect.DeepEqual(statRet, services.s) {
+	if marshalOrDie(t, statRet) != marshalOrDie(t, services.s) {
 		t.Error("Error translating FromAPIService")
 	}
 }
+
+func marshalOrDie(t *testing.T, v interface{}) string {
```
stephenrlouie: NIT. [marshallVsTheMachines](https://youtu.be/QIjgTutJdpk) is the proper name

api/v1/server/server_test.go
```diff
@@ -501,33 +495,29 @@ var nilService = apiclient.Service{
...
 			Value: []string{"Mike", "adam"},
 			From:  9,
 		},
-		apiclient.ServiceIo{
-			Name:  "in3",
-			Type_: "in3",
-			Value: 3.0,
-			From:  "adam",
-		},
+		//		"in3": apiclient.ServiceIo{
```

api/v1/convert/convert_test.go
adamschaub: How about `marshalOrGTFO`?

Issue #116, opened 2017-06-01
Service idempotency and queue back pressure


adamschaub:
Dual issue, could break this down into two separate cases.

Currently, if the `service init interval` (interval to queue services to be updated) is too small, services create get queued for update faster than they get processed. Run sigma with these flags and log into the rabbitmq console to see the queue slowly bloat:

```
$ ./bin/sigma --config config.yaml --manager-service-init-interval 500ms --manager-operation-status-interval 10ms --manager-update-interval 10ms
```

Then the issue arises that our updates are truly idempotent:
```
ERRO[2017-06-01T11:21:50-04:00] Could not start operation 98f555f1-2256-406e-92cd-e3d25a12e726-operation  error="Error response from daemon: Conflict. The container name "/98f555f1-2256-406e-92cd-e3d25a12e726" is already in use by container "e19f8a8393b8716497bd5b59c738ed3cfb2ee59f63005a60e0e9b370a47db55d". You have to remove (or rename) that container to be able to reuse that name." file=manager.go line=381
ERRO[2017-06-01T11:21:50-04:00] Unable to update service operation 98f555f1-2256-406e-92cd-e3d25a12e726-operation  error="Key already exists" file=manager.go line=511
ERRO[2017-06-01T11:21:50-04:00] Could not get status of 98f555f1-2256-406e-92cd-e3d25a12e726-operation  error="Error: No such container: 98f555f1-2256-406e-92cd-e3d25a12e726" file=manager.go line=401
ERRO[2017-06-01T11:21:51-04:00] Unable to cleanup operation 98f555f1-2256-406e-92cd-e3d25a12e726-operation  error="Error response from daemon: No such container: 98f555f1-2256-406e-92cd-e3d25a12e726" file=manager.go line=494
```

The case here is due to docker tasker not appropriately returning a `ErrServiceAlreadyRunning` error when multiple requests to run the same container are issued: https://github.com/sigma-dev/sigma/blob/master/manager/manager.go#L376

mcaulfield:
Second part addressed by #118. First part needs a check for max backlog size to avoid overwhelming message queue.

PR #118, opened 2017-06-01
Taskers to return AlreadyRunning error

adamschaub:
Partial point from #116 

Tasker `Run` method must check if the given job is already running, and return no error if true. 

- [x] Docker tasker


meganokeefe:
a note that we didn't have to actually change nomadTasker for this, since re-adding an existing Job is a no-op / ignored by Nomad.

adamschaub:
@meganokeefe Yeah, as per our conversation on Friday. We determined that a "failure" to start another job using the tasker because it already exists should constitute as "no error".

What `ErrIsNoLongerRunning`? I haven't removed any of those.

PR #119, opened 2017-06-01
Service.Site hot fixes

meganokeefe:
1. AddService: we make sure the provided site exists in the DB. If it doesn't, we add the site. 
2. UpdateService: if the site is empty, we default to `site-local` (this logic already in AddService). Ans also makes sure the provided site exists in the DB / if it doesn't, we add it. 
3. Shoehorning into this PR: preventing users from deleting site=`local`.


api/v1/server/server.go
```diff
@@ -216,6 +216,16 @@ func (s *Server) AddService(w http.ResponseWriter, r *http.Request) {
 	modObj := convert.FromAPIService(apiObj)
 	if len(modObj.Site) == 0 {
 		modObj.Site = model.DefaultSite
+	} else {
+		_, err := s.db.GetSite(modObj.Site)
+		// insert the provided site into the DB if not exists
```
adamschaub: do we really want to insert the site? Or return an error instead. Seems to me that sites would be installed separately.

api/v1/server/server.go
```diff
@@ -759,8 +773,17 @@ func (s *Server) UpdateService(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if len(apiObj.Site) == 0 {
-		http.Error(w, "Site is required in a Service", http.StatusBadRequest)
-		return
+		apiObj.Site = model.DefaultSite
+	} else {
+		_, err := s.db.GetSite(apiObj.Site)
+		// insert the provided site into the DB if not exists
```
adamschaub: Same Q as above. Also in direct violation of DRY

PR #121, opened 2017-06-01
Auth

mikeynap:
Framework for authorization.
auth has a Tokenizer (JWT) and an authorizer (will be LDAP or DB etc).

Authorizer takes username/password and validates.
Tokenizer creates a token.

Added /v1/login endpoint:
Given username, password returns Token.

This Token is used in all requests. There is middleware around all the calls that requires the token set.

Currently errthing is set to the  null versions, so login() is not required and the endpoints are unprotected.


api/v1/client/token.go
```diff
@@ -0,0 +1,16 @@
...
+ * OpenAPI spec version: 1.0
+ *
+ * Generated by: https://github.com/swagger-api/swagger-codegen.git
+ */
+
+package apiclient
+
+type Token struct {
+	Token   string `json:"token,omitempty"`
+	Expires int64  `json:"duration,omitempty"`
```
meganokeefe: how is `Expires` expressed? seconds, milliseconds?

api/v1/gen/v1.yaml
```diff
@@ -56,6 +74,18 @@ parameters:
...
 # Describe your paths here
 paths:
+  /login:
+    get:
+      security:
+        - Basic: []
+      description: generate token using basic auth.
+      operationId: loginWithStoredUsernameAndPassword
+      responses:
+        200:
```
meganokeefe: shouldn't it be possible to have a non-200 response here?

api/v1/server/server.go
```diff
@@ -180,24 +188,182 @@ func (s *Server) Stop() {
...
+	if err != nil {
+		log.Error(err)
+		w.Header().Set("WWW-Authenticate", `Basic realm="Enter Login for API Token"`)
+		http.Error(w, "Bad Login", http.StatusUnauthorized)
+		return
+	}
+	apiTok := convert.FromAuthToken(tok)
+	w.WriteHeader(http.StatusOK)
+	if err = json.NewEncoder(w).Encode(apiTok); err != nil {
+		log.Errorf("Unable to Encode Token: %v", err)
```
meganokeefe: do we want to return an http error here? bad request?

api/v1/server/server.go
```diff
@@ -180,24 +188,182 @@ func (s *Server) Stop() {
...
+	if users, err = s.auth.GetUsernames(nil); err != nil {
+		if err == user.ErrNotAllowed {
+			http.Error(w, err.Error(), http.StatusUnauthorized)
+			return
+		}
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	if users == nil {
+		users = make([]string, 0)
```
meganokeefe: what is this for?

api/v1/server/server.go
```diff
@@ -180,24 +188,182 @@ func (s *Server) Stop() {
...
+
 	modObj := convert.FromAPIPlugin(apiObj)
 	var theID string
 	var err error
 	if theID, err = s.db.InsertPlugin(modObj); err != nil {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
-	w.Header().Add("Location", fmt.Sprintf("/v1/services/%v", theID))
+	w.Header().Add("Location", fmt.Sprintf("/v1/plugins/%v", theID))
```
meganokeefe: good catch

auth/authorizer/db.go
```diff
@@ -0,0 +1,153 @@
+package authorizer
```
meganokeefe: unit tests for these functions plz?

auth/authorizer/db.go
```diff
@@ -0,0 +1,153 @@
...
+
+	if group != "" {
+		su.AuthGroup = group
+	}
+
+	return a.sdb.UpdateSecret(su)
+}
+
+func (a *DBAuth) GetUser(un string) (user.User, error) {
+	u, err := a.sdb.GetSecret(newSecretUser(un))
```
meganokeefe: is this just-in-time struct creation for DB insertion idiomatic?
You're converting to-and-from here. 

Also, why is Secret an interface in `database.go`, vs a struct in `model.Go`?

auth/authorizer/db.go
```diff
@@ -0,0 +1,153 @@
...
+
+	var suser *secretUser
+	var ok bool
+	if suser, ok = u.(*secretUser); !ok || suser == nil || suser.Id == "" {
+		return nil, ErrBadDatabaseResults
+	}
+	return suser.AuthUser, nil
+}
+
+func (a *DBAuth) GetUsernames(filters []string) ([]string, error) {
```
meganokeefe: are you documenting what filters should look like anywhere? 

is it a list like `["x=x_val", "y=y_val"]`?

auth/authorizer/db.go
```diff
@@ -0,0 +1,153 @@
...
+	filtMap := map[string]string{}
+	for _, f := range filters {
+		parts := strings.Split(f, "=")
+		if len(parts) != 2 {
```
meganokeefe: you're not rejecting invalid filters? do you want to `log.Warn` or anything here?

auth/authorizer/db.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	users, err := a.sdb.QuerySecrets(newSecretUser(""), filtMap)
+	if err != nil {
+		return nil, err
+	}
+	return users, nil
+}
+
+func (a *DBAuth) Login(un, pw string) (user.User, error) {
+	u, err := a.sdb.GetSecret(newSecretUser(un))
```
meganokeefe: see comment about typing above

auth/tokenizer/jwt.go
```diff
@@ -0,0 +1,84 @@
+package tokenizer
```
meganokeefe: unit tests please

database/couchbase.go
```diff
@@ -222,7 +222,7 @@ func (cb *Couchbase) IdFromName(n string, oType string) string {
 	if n == "" {
 		return ""
 	}
-	rows, err := cb.executeQuery("SELECT Id FROM `"+cb.cfg.BucketName+"` WHERE Name=$name AND Metadata.Type=$type", map[string]interface{}{
+	rows, err := cb.executeQuery("SELECT Id FROM `"+cb.cfg.BucketName+"` WHERE (Name=$name OR Metadata.Name=$name) AND Metadata.Type=$type", map[string]interface{}{
```
meganokeefe: why this change?

database/couchbase.go
```diff
@@ -609,12 +612,8 @@ func isClean(s string) bool {
 	return reg.MatchString(s)
 }
 
-func (cb Couchbase) makeQuery(obj model.Metadataer, idOnly bool) (string, map[string]interface{}, error) {
+func (cb Couchbase) makeQuery(obj model.Metadataer, selector string, filters map[string]string) (string, map[string]interface{}, error) {
```
meganokeefe: need to add Secret query info to the switch in `makeQuery`?

debug.yaml
```diff
@@ -0,0 +1,17 @@
+---
```
meganokeefe: should this config be merged into master?

lol `jitsNOTnits`

api/v1/server/server.go
mikeynap: Unfortunately you cannot -- Once you write to (w) you cannot write anymore headers. 

This is how we do it in the whole project, if we wanna try something else we'll do it in a different PR.

api/v1/client/token.go
mikeynap: It's a unix date.

api/v1/server/server.go
mikeynap: So that JSON.Encode writes [] and not an empty body.

auth/authorizer/db.go
mikeynap: We'll chat about this --

auth/authorizer/db.go
mikeynap: yes, I'll document.

debug.yaml
mikeynap: No it shouldn't -- I'll git rm.

PR #123, opened 2017-06-02
No Update on Service put

adamschaub:
- [x] Remove "Dirty" metadata from documents. Rely only on changed inputs/outputs for triggering service updates. 
- [x] Fix ordering issues in state handling for enable & remove. 
  - [x] Enable set to false destroys the service (if active)
  - [x] Remove set to true deletes the service from db (if inactive or destroyed)
- [x] Update tests & replace nulldb with moc


manager/manager_test.go
```diff
@@ -493,22 +484,21 @@ func TestUpdateService(t *testing.T) {
...
-				PluginName:    "a",
-				PluginVersion: "b",
-				Status:        model.ServiceStatusInactive,
-				Site:          "testsite",
-				Enabled:       model.NewBool(true),
-			},
-			expectStatus: model.ServiceStatusActive,
-			plugin:       model.Plugin{},
-		},
+	}{ /*
```
vgunning: Dead code?

adamschaub: WHOOOPS

adamschaub: fixed

meganokeefe:
Reviewed + LGTM.

PR #125, opened 2017-06-05
Embeded Config

mikeynap:
Embedded Structs are allowed in Config now. Example:
```
Config{
     Auth: {
                   LDAP{
                                 ASDF: true
                   },
     },
}
```


config/config.go
```diff
@@ -111,12 +116,11 @@ func getCfg(gCfg *GlobalConfig) error {
...
 			str = flagStr
 		}
-
-		if len(str) != 0 {
+		if len(str) != 0 || subField.Type().Kind() == reflect.Struct {
 			switch subField.Type().Kind() {
 			case reflect.Bool:
 				v := viper.GetBool(str)
-				subField.SetBool(v)
+				subField.SetBool(v || subField.Bool()) // IsSet is broken with bools, see NOTE above ^^^
```
adamschaub: works

adamschaub:
I'm floored that this wasn't more invasive.

adamschaub:
We need to set up some bread crumbs to properly create flag/env var names and make sure top level struct/field names are preserved.

Example adding a nested struct to RabbitConfig:
```
type RabbitConfig struct {
	Username string
	Password string
	URL      string
	Nested   struct {
		I int `desc:"nested struct in Rabbit"`
	}
}
```

Generates the flags:
```
      --nested-i int                               nested struct in Rabbit
      --rabbit-password string                     
      --rabbit-url string                          
      --rabbit-username string   
```


Ideally the flag should be `--rabbit-nested-i`.

config/config.go
```diff
@@ -88,35 +79,36 @@ func GetConfig() (GlobalConfig, error) {
...
-	return eachSubField(gCfg, func(parent reflect.Value, fieldName string, subFieldName string) error {
-		envStr := envString(fieldName, subFieldName)
-		flagStr := flagString(fieldName, subFieldName)
+	return eachSubField(gCfg, func(parent reflect.Value, subFieldName string, crumbs []string) error {
+		p := strings.Join(crumbs, "")
+		envStr := envString(p, subFieldName)
+		flagStr := flagString(p, subFieldName)
 
-		// eachSubField only calls this function if both fieldName and subFieldName exist
+		// eachSubField only calls this function if  subFieldName exists
```
adamschaub: NIT: extra space

(jk, leave it)

PR #128, opened 2017-06-06
Vet on declared type

adamschaub:
Ref #98

- [x] Reject Plugins/Services with invalid IO types
  - [x] Plugin must exist before service is created (if service isn't bundle)
- [x] All tests/examples use proper types
- [x] Plugin types with optional field
- [x] If output types don't match, attempt convert
  - [ ] Only convert sometimes? Maybe a strict flag on plugin types?
- [x] Make sure any errors are visible (StatusDescription?)


api/v1/server/server.go
```diff
@@ -226,6 +242,38 @@ func (s *Server) AddService(w http.ResponseWriter, r *http.Request) {
...
+				http.Error(w, fmt.Sprintf("No matching plugin"), http.StatusBadRequest)
+				return
+			}
+			plugin = plugins[0]
+		} else {
+			http.Error(w, fmt.Sprintf("Unable to verify plugin: %s", err.Error()), http.StatusInternalServerError)
+		}
+	}
+
+	if errs := model.VetServiceIO(modObj, plugin); len(errs) > 0 {
```
meganokeefe: "don't repeat yourself?"

api/v1/server/server.go
```diff
@@ -191,6 +191,22 @@ func (s *Server) AddPlugin(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	modObj := convert.FromAPIPlugin(apiObj)
+
+	if errs := model.VetPluginIO(modObj); len(errs) > 0 {
+		allErrs := struct {
```
meganokeefe: why not have `VetPluginIO` do this structing, marshalling logic?

model/convert.go
```diff
@@ -0,0 +1,189 @@
...
+		// If a isn't directly assertable to object type, can't be done
+		if obj, ok := a.(map[string]interface{}); ok {
+			return obj, true
+		}
+		return map[string]interface{}{}, false
+	case TypeInt:
+		switch v := a.(type) {
+		case int:
+			return v, true
+		case float32:
```
meganokeefe: can an `Int` be a `float`?

model/convert.go
```diff
@@ -0,0 +1,189 @@
...
+			}
+
+			var l []interface{}
+
+			for i := 0; i < typ.Len(); i++ {
+				ele := typ.Index(i)
+				if v, ok := convertBasicIO(ele.Interface(), out); ok {
+					l = append(l, v)
+				} else {
+					return listOfType(out), false
```
meganokeefe: reason behind having both `listOfType()` and `toListOfType()`?

model/convert.go
```diff
@@ -0,0 +1,189 @@
...
+			if final := toListOfType(l, out); final != nil {
+				return final, true
+			}
+		}
+		return nil, false
+	}
+}
+
+func convertPluginIO(a interface{}, t string, plug Plugin) (interface{}, bool) {
+
```
meganokeefe: nit remove whitespace

model/convert.go
```diff
@@ -0,0 +1,189 @@
...
+				return nil
+			}
+		}
+		return a
+	}
+	return nil
+}
+
+// returns an empty list of the given basic type
+func listOfType(t string) interface{} {
```
meganokeefe: nit- rename to `EmptyListOfType`?

model/model.go
```diff
@@ -217,33 +189,9 @@ func ServiceDataItemsToMap(in map[string]ServiceDataItem) map[string]interface{}
 	return sdimap
 }
 
-// CanonicalizeOutput goes from the generic output to our
-// specific map[string][]string format.
-// Do this until we change to a more complex type system for outputs.
-func CanonicalizeOutput(m map[string]interface{}) map[string][]string {
```
meganokeefe: I don't see this re-added anywhere, are your changes here a replacement for `Canonicalize?`

model/vet.go
```diff
@@ -0,0 +1,72 @@
+package model
```
meganokeefe: nit, rename `vet` to `validate`? feel like that's more standard

model/vet.go
```diff
@@ -0,0 +1,72 @@
...
+
+	return
+}
+
+func ParseFrom(f string) (string, string, bool) {
+	parts := strings.Split(f, ".")
+	if len(parts) < 2 {
+		return "", "", false
+	}
+	svcname, varname := strings.Join(parts[0:len(parts)-1], "."), parts[len(parts)-1]
```
meganokeefe: are we going to trust users to not have periods in their variable names? because that would screw this up

api/v1/server/server.go
adamschaub: VetPluginIO just returns a list of errors. This is packaging it to send back through the API.

model/convert.go
adamschaub: I can probably clean this up a bit to remove `listOfType`. As it stands, they serve different purposes. `listOfType` creates an empty list given the basic type. `toListOfType` converts a `[]interface{}` to given type.

model/convert.go
adamschaub: no.

model/convert.go
adamschaub: Sure. But I might just remove in refactoring. I'll make note

model/model.go
adamschaub: `Canonicalize` wasn't being used anywhere, so I removed it.

model/vet.go
adamschaub: Sure

model/vet.go
adamschaub: This is Mike's code, and I'm pretty sure he did doc it. I'll check on it

model/vet.go
mikeynap: I would say vet is more "Go standard"

vgunning:
Should the `api/v1/convert/convert_test.go` be updated to include `apiclient.TypeEntry` in the plugin conversion tests?

model/convert.go
```diff
@@ -0,0 +1,275 @@
...
+			return v, true
+		case float64:
+			return v, true
+		case string:
+			f, err := strconv.ParseFloat(v, 64)
+			if err == nil {
+				return f, true
+			}
+		}
+		return float64(0), false
```
vgunning: are we defaulting to the `float64` type?
if so, should we `return float64(v), true` on the `float32` and `float64` cases

adamschaub: Good point. I'm not entirely sure. I felt that if the type was float32, we might as well keep it consistent and avoid introducing errors when upcasting

model/metadata.go
```diff
@@ -0,0 +1,104 @@
...
+//GetString returns the string value of metadata at key k.
+// If the underlying type isn't string, a conversion is attempted.
+// Returns false if the key doesn't exist, or the value is not
+// able to be converted to a string
+func (m *MetaImpl) GetString(k string) (string, bool) {
+	if v, ok := m.Get(k); ok {
+		switch v := v.(type) {
+		case int:
+			return fmt.Sprintf("%d", v), true
+		case int64:
```
vgunning: reason that int64 is suddenly included?

model/metadata.go
```diff
@@ -0,0 +1,104 @@
...
+// GetInt returns the integer value of metadata at key k.
+// If the underlying type isn't int, a conversion is attempted.
+// Returns false if the key doesn't exist, or the value is not able
+// to be converted to an int
+//
+// Integer value may be overflowed if the underlying type is float32/64 or int64
+func (m *MetaImpl) GetInt(k string) (int, bool) {
+	if v, ok := m.Get(k); ok {
+		switch v := v.(type) {
+		case int:
```
vgunning: NIT: would condensing improve readability?
`case int, int8, int16, int32, int64: return int(v), true` etc...
also, lack of `int32`?

model/metadata.go
adamschaub: I don't follow

model/metadata.go
adamschaub: You cannot condense a type switch, I will add int32

model/convert.go
```diff
@@ -0,0 +1,275 @@
...
+import (
+	"reflect"
+	"strconv"
+)
+
+// ConvertIO returns the value a as the provided type t. If t is
+// not a basic type, plug is checked for an appropriate type definition
+// If conversion isn't possible, a zero value of the given type is returned
+func ConvertIO(a interface{}, t string, plug Plugin) (interface{}, bool) {
+	if out, ok := IsList(t); ok {
```
mikeynap: needs more levels of indentation

model/convert.go
```diff
@@ -0,0 +1,275 @@
...
+	if !ok {
+		return nil, false
+	}
+
+	// Eventually typ.Base will be a "basic" type
+	out, ok := ConvertIO(a, typ.Base, plug)
+
+	// If type is object, convert subfields
+	failed := !ok
+	if typ.Base == TypeObject {
```
mikeynap: wanna be a pal and remove these first two blocks (return if either is not True).

model/convert.go
```diff
@@ -0,0 +1,275 @@
...
+		a := []string{}
+		for _, v := range l {
+			if b, ok := v.(string); ok {
+				a = append(a, b)
+			} else {
+				return nil
+			}
+		}
+		return a
+	default:
```
mikeynap: remove this from default.

model/convert.go
adamschaub: RIP

model/convert.go
adamschaub: GOOD point

model/convert.go
adamschaub: word

PR #136, opened 2017-06-08
Glues LDAP + Auth2

mikeynap:
To Be Merged after Both LDAP + Auth2


api/v1/convert/convert.go
```diff
@@ -229,3 +231,24 @@ func FromModelSite(o model.Site) apiclient.Site {
...
+	if o == nil {
+		return apiclient.Token{}
+	}
+	return apiclient.Token{
+		Token:   o.String(),
+		Expires: o.Expires(),
+	}
+}
+
+func FromAuthUser(o user.User) apiclient.User {
```
adamschaub: Where's the tests for this kappa

api/v1/server/middleware.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"net/http"
+	"strings"
+
+	"github.com/sigma-dev/sigma/auth/user"
+
+	"golang.org/x/net/context"
+)
+
+const CTX_AUTH_USER = "CTX_AUTH_USER"
```
adamschaub: More standard way is to define an unexported type as the key. Ensures no key conflicts in the same context.

See `Package userip`: https://blog.golang.org/context

api/v1/server/routers.go
```diff
@@ -48,169 +48,213 @@ func (r *Router) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+			Handler:    s.Login,
+			Middleware: []Middleware{},
+		},
+
+		Route{
+			Name:       "AddUser",
+			Method:     "POST",
+			Pattern:    "/v1/secrets/users",
+			Handler:    s.AddUser,
+			Middleware: []Middleware{s.AuthUserAdmin, s.Auth},
```
adamschaub: Does this order work? Based on how Auth/AuthUserAdmin work, Auth must happen first. Is the list of middleware treated like a stack?

api/v1/server/server.go
```diff
@@ -180,24 +187,191 @@ func (s *Server) Stop() {
...
+	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+	un, pw, ok := r.BasicAuth()
+	if !ok {
+		w.Header().Set("WWW-Authenticate", `Basic realm="Enter Login for API Token"`)
+		http.Error(w, "Not Authorized", http.StatusUnauthorized)
+		return
+	}
+	tok, err := s.auth.LoginAndTokenize(un, pw)
+	if err != nil {
+		log.Error(err)
```
adamschaub: `WithError` and include some context in the log line

api/v1/server/server.go
```diff
@@ -180,24 +187,191 @@ func (s *Server) Stop() {
...
+		w.Header().Set("WWW-Authenticate", `Basic realm="Enter Login for API Token"`)
+		http.Error(w, "Bad Login", http.StatusUnauthorized)
+		return
+	}
+	apiTok := convert.FromAuthToken(tok)
+	w.WriteHeader(http.StatusOK)
+	if err = json.NewEncoder(w).Encode(apiTok); err != nil {
+		log.Errorf("Unable to Encode Token: %v", err)
```
adamschaub: `WithError`

api/v1/server/server.go
```diff
@@ -180,24 +187,191 @@ func (s *Server) Stop() {
...
+	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+	decoder := json.NewDecoder(r.Body)
+	defer r.Body.Close()
+	var u apiclient.User
+	if err := decoder.Decode(&u); err != nil || u.Name == "" {
+		http.Error(w, fmt.Sprintf("Invalid User Provided: %v", err), http.StatusBadRequest)
+		return
+	}
+	if err := s.auth.AddUser(u.Name, u.Password, user.Role(u.Role)); err != nil {
+		if err.Error() == auth.ErrOperationNotImplemented.Error() {
```
adamschaub: Why the string comparison if `ErrOperationNotImplemented` is a constant?

api/v1/server/routers.go
mikeynap: Yeah it's treated as a stack -- I should probably reverse it for clarity though shouldn't I...

api/v1/server/middleware.go
mikeynap: cool, I'll use that.

auth/authorizer/db.go
```diff
@@ -0,0 +1,171 @@
...
+		if err != nil {
+			log.WithError(err).Errorf("Unable to create default user `admin`")
+		} else {
+			log.Info("Created default user `admin` with password `admin`.")
+		}
+	}
+	return a
+}
+
+func (a *DBAuth) AddUser(un, pw string, role user.Role) error {
```
adamschaub: Need to do some validation on username as well. Check that it has no whitespace characters, at a minimum.

api/v1/server/server.go
mikeynap: Not entirely sure every bad login should be logged to Error... maybe debug.

api/v1/server/server.go
```diff
@@ -180,24 +187,191 @@ func (s *Server) Stop() {
...
+			http.Error(w, err.Error(), http.StatusNotImplemented)
+			return
+		}
+		http.Error(w, err.Error(), http.StatusBadRequest)
```
adamschaub: Also have the potential to be an auth DB error. Might want to differentiate bad request/internal error

api/v1/server/server.go
```diff
@@ -180,24 +187,191 @@ func (s *Server) Stop() {
...
+			return
+		} else if err.Error() == auth.ErrOperationNotImplemented.Error() {
+			http.Error(w, err.Error(), http.StatusNotImplemented)
+			return
+		}
+
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	if users == nil {
```
adamschaub: Really makes you yearn for a ternary operator

api/v1/server/server.go
```diff
@@ -180,24 +187,191 @@ func (s *Server) Stop() {
...
+		users = make([]string, 0)
+	}
+
+	w.WriteHeader(http.StatusOK)
+	if err = json.NewEncoder(w).Encode(users); err != nil {
+		log.Errorf("Unable to Encode Users: %v", err)
+		return
+	}
+	log.Debugf("Users GET")
```

api/v1/server/server.go
mikeynap: Yeah I fixed that -- my bad

api/v1/server/server.go
```diff
@@ -180,24 +187,191 @@ func (s *Server) Stop() {
...
+	defer r.Body.Close()
+	var apiObj apiclient.User
+	if err := decoder.Decode(&apiObj); err != nil || apiObj.Name == "" {
+		http.Error(w, fmt.Sprintf("Invalid User Name Provided: %v", err), http.StatusBadRequest)
+		return
+	}
+
+	updateFunc := func() (int, error) {
+		if err := s.auth.UpdateUser(apiObj.Name, apiObj.Password, apiObj.Group, user.Role(apiObj.Role)); err != nil {
+			log.Errorf("Update err: %v", err)
```

auth/authorizer/db.go
mikeynap: moved to the appropriate auth function (db.go). (since, for example, ldap would allow special chars etc).

api/v1/server/server.go
```diff
@@ -180,24 +187,191 @@ func (s *Server) Stop() {
...
+
 	modObj := convert.FromAPIPlugin(apiObj)
 	var theID string
 	var err error
 	if theID, err = s.db.InsertPlugin(modObj); err != nil {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
-	w.Header().Add("Location", fmt.Sprintf("/v1/services/%v", theID))
+	w.Header().Add("Location", fmt.Sprintf("/v1/plugins/%v", theID))
```
adamschaub: good call

api/v1/server/server_test.go
```diff
@@ -608,6 +612,30 @@ func populate() {
...
+		t.Error("Should have not been unauthorized.")
+	}
+	ot := server.auth.Tokenizer
+	server.auth.Tokenizer = tokenizer.NewJWT("blah")
+	r, _ = client.AddPlugin(filledInPlugin)
+	if r.StatusCode != http.StatusUnauthorized {
+		t.Error("Should have been unauthorized.")
+	}
+
+	res, a, b := client.Login("admin", "admin")
```
adamschaub: If we can avoid having these hard coded here, that would be great. Either find a way to add the user from the test, or have a constant defined that can be changed in one place

api/v1/server/server_test.go
```diff
@@ -608,6 +612,30 @@ func populate() {
...
+	if res == nil || res.Token == "" || res.Expires == 0 {
+		t.Error("Login Should have not returned nil.", res, string(a.Payload), b)
+	}
+
+	r, _ = client.AddPlugin(filledInPlugin)
+	if r.StatusCode == http.StatusUnauthorized {
+		t.Error("Should have NOT been unauthorized.")
+	}
+
+	server.auth.Tokenizer = ot
```
adamschaub: what's the point?

auth/auth.go
```diff
@@ -0,0 +1,53 @@
...
+	LDAP         authorizer.LDAPConfig
+}
+
+func NewConfig() Config {
+	return Config{
+		Method: "",
+	}
+}
+
+type Auth interface {
```
adamschaub: Me gusta

api/v1/server/server_test.go
```diff
@@ -42,7 +45,9 @@ func TestMain(m *testing.M) {
 		db = database.NewNullDB(false)
```
adamschaub: Why is server_test still using nullDB and not a mock? We should be able to remove a lot of this setup and focus on actual unit tests. Different day, different PR though...

auth/auth.go
```diff
@@ -0,0 +1,53 @@
...
+	tokenizer.Tokenizer
+	LoginAndTokenize(un, pw string) (tokenizer.Token, error)
+}
+
+type AuthStruct struct {
+	authorizer.Authorizer
+	tokenizer.Tokenizer
+}
+
+func (a AuthStruct) LoginAndTokenize(un, pw string) (tokenizer.Token, error) {
```
adamschaub: Tests for this? kappa

auth/authorizer/db.go
```diff
@@ -0,0 +1,177 @@
...
+	"unicode"
+
+	"github.com/pborman/uuid"
+	"github.com/sigma-dev/sigma/auth/user"
+	"github.com/sigma-dev/sigma/database"
+	"github.com/sigma-dev/sigma/log"
+	"github.com/sigma-dev/sigma/model"
+)
+
+const dbType = "secretUser"
```
adamschaub: Not used anywhere

auth/authorizer/db_test.go
```diff
@@ -0,0 +1,15 @@
+package authorizer
+
+import "testing"
+
+func TestPasswordStuff(t *testing.T) {
+	u := &secretUser{}
+	p := "s0m∑p4zzw0rd"
```
adamschaub: Password can only contain alpha-characters and cannot be longer than 8 characters long

auth/authorizer/db.go
```diff
@@ -0,0 +1,177 @@
...
+func newSecretUser(n string) *secretUser {
+	return &secretUser{
+		AuthUser: user.AuthUser{
+			AuthName: n,
+		},
+	}
+}
+
+func (s secretUser) Type() string {
+	return "secretUser"
```
adamschaub: `return dbType`

auth/authorizer/db.go
adamschaub: ldap would allow, but we use this same user name in URLs for the API server. LCD means we have to put some restrictions on it.

auth/authorizer/db.go
```diff
@@ -0,0 +1,177 @@
...
+	return a
+}
+
+func (a *DBAuth) AddUser(un, pw string, role user.Role) error {
+	if len(pw) == 0 {
+		return ErrBadPassword
+	}
+
+	for _, l := range un {
+		if unicode.IsSymbol(l) || unicode.IsSpace(l) {
```
adamschaub: Danke

auth/authorizer/db.go
```diff
@@ -0,0 +1,177 @@
...
+	return a.sdb.DeleteSecret(newSecretUser(un))
+}
+
+func (a *DBAuth) UpdateUser(un, pw, group string, role user.Role) error {
+	s, err := a.sdb.GetSecret(newSecretUser(un))
+	if err != nil {
+		return err
+	}
+	su, ok := s.(*secretUser)
+	if !ok || su.Id == "" {
```
adamschaub: You check `su != nil` below in `GetUser` after calling `sdb.GetSecret`. Is that a check that should be done here as well, is it possible for `GetSecret` to return `nil`? Lest `su.Id` panics.

auth/authorizer/db.go
```diff
@@ -0,0 +1,177 @@
...
+func (a *DBAuth) GetUser(un string) (user.User, error) {
+	u, err := a.sdb.GetSecret(newSecretUser(un))
+	if err != nil {
+		return nil, ErrBadDatabaseResults
+	}
+
+	var suser *secretUser
+	var ok bool
+	if suser, ok = u.(*secretUser); !ok || suser == nil || suser.Id == "" {
+		return nil, ErrBadDatabaseResults
```
adamschaub: Differentiate `BadDatabaseResults` and `NotFound`?

auth/authorizer/db.go
```diff
@@ -0,0 +1,177 @@
...
+func (a *DBAuth) Login(un, pw string) (user.User, error) {
+	u, err := a.sdb.GetSecret(newSecretUser(un))
+	if err != nil {
+		return nil, ErrBadLogin
+	}
+
+	var suser *secretUser
+	var ok bool
+	if suser, ok = u.(*secretUser); !ok || suser == nil || suser.Id == "" {
+		return nil, ErrBadLogin
```
adamschaub: `BadLogin` or database error?

auth/authorizer/ldap.go
```diff
@@ -0,0 +1,304 @@
...
+
+type LDAPUser struct {
+	user.AuthUser
+	DistinguishedName string
+	Domain            string
+}
+
+var (
+	ErrBadAdminCreds = errors.New("bad admin credentials - please use fully distinguished usernames")
+	ErrNotExist      = errors.New("queried use does not exist")
```
adamschaub: `queried use` -> `queried user`

auth/authorizer/ldap.go
```diff
@@ -0,0 +1,304 @@
...
+type LDAPConfig struct {
+	Host                   string
+	UseTLS                 bool `flag:"false"`
+	InsecureSkipVerify     bool `flag:"false"`
+	AdminDistinguishedName string
+	AdminPassword          string
+
+	DefaultUserDomain              string `flag:"false"`
+	DefaultUserOrganizationalUnits string `flag:"false"`
+}
```
adamschaub: Why not flags for some of these? I could see `UseTLS` and `InsecureSkipVerify` as useful flags

auth/authorizer/ldap.go
```diff
@@ -0,0 +1,304 @@
...
+	dcs []string
+}
+
+type LDAPUser struct {
+	user.AuthUser
+	DistinguishedName string
+	Domain            string
+}
+
+var (
```
adamschaub: move this declaration to the top

auth/authorizer/ldap.go
```diff
@@ -0,0 +1,304 @@
...
+
+func (a *LDAPAuth) UpdateUser(un, pw, group string, role user.Role) error {
+	return ErrOperationNotImplemented
+}
+
+func (a *LDAPAuth) GetUser(un string) (user.User, error) {
+	if a.AdminDistinguishedName == "" {
+		return nil, user.ErrNotAllowed
+	}
+	l, err := a.dialAndBind(a.AdminDistinguishedName, a.AdminPassword)
```
adamschaub: Do we really dial every time we get user? Wouldn't it be better to create the connection once and reuse?

auth/authorizer/db.go
mikeynap: oops. type now returns it

auth/authorizer/db_test.go
mikeynap: ??? why???

auth/authorizer/db.go
mikeynap: ya

auth/authorizer/ldap.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+func (a *LDAPAuth) Login(un, pw string) (user.User, error) {
+	if !strings.Contains(strings.ToLower(un), "dc=") && strings.Contains(strings.ToLower(a.DefaultUserDomain), "dc=") {
+		un = fmt.Sprintf("cn=%s,%s", un, a.DefaultUserDomain)
+	}
+
+	l, err := a.dialAndBind(un, pw)
+	if err != nil {
+		un = strings.Replace(un, "cn=", "uid=", -1)
```
adamschaub: What's up with this replacement. If it's something that will cause dialAndBind to fail, why not just do it before the first attempt?

auth/authorizer/ldap.go
adamschaub: If no to that, should `l.Close()` be called?

auth/authorizer/ldap.go
mikeynap: I guess..,

auth/authorizer/ldap.go
```diff
@@ -0,0 +1,304 @@
...
+			AuthGroup: "",
+			AuthRole:  role,
+		},
+	}, nil
+}
+
+// converts domain name (a.b.com) into LDAP format
+// dc=a,dc=b,dc=com
+func domainToDC(domain string) (string, []string) {
+	domain = strings.Replace(domain, " ", "", -1)
```
adamschaub: Isn't the domain name just invalid if it has spaces in it?

auth/authorizer/ldap.go
```diff
@@ -0,0 +1,304 @@
...
+		return domain, nil
+	}
+	parts := strings.Split(domain, ".")
+	r := ""
+	var dcs []string
+	for _, p := range parts {
+		r += fmt.Sprintf("dc=%s,", p)
+		dcs = append(dcs, fmt.Sprintf("dc=%s", p))
+	}
+	return strings.TrimSuffix(r, ","), dcs
```
adamschaub: Instead of doing this trim, you could just create the list of strings, then do a `strings.Join` on dcs.

auth/authorizer/ldap.go
```diff
@@ -0,0 +1,304 @@
...
+	var ous []string
+	for _, p := range parts {
+		p = strings.TrimSpace(p)
+		if !strings.HasPrefix(strings.ToLower(p), "ou=") {
+			p = "ou=" + p
+		}
+		r += p + ","
+		ous = append(ous, p)
+	}
+	return strings.TrimSuffix(r, ","), ous
```
adamschaub: Same as above. Just `strings.Join` on `ous`.

auth/authorizer/ldap.go
```diff
@@ -0,0 +1,304 @@
...
+				a.ous = append(a.ous, p)
+			}
+			if a.DefaultUserDomain == "" && strings.HasPrefix(strings.TrimSpace(strings.ToLower(p)), "dc=") {
+				domain += p + ","
+				a.dcs = append(a.dcs, p)
+			}
+		}
+
+		a.DefaultUserOrganizationalUnits = strings.TrimSuffix(ou, ",")
+		a.DefaultUserDomain = strings.TrimSuffix(domain, ",")
```
adamschaub: `strings.Join`

auth/authorizer/ldap.go
mikeynap: This is for the first part -- if it's in the form cn=admin,dc=blah, dc=blah2.

auth/authorizer/ldap.go
mikeynap: good call

auth/authorizer/ldap.sh
```diff
@@ -0,0 +1,43 @@
...
+
+if [ $c = "stop" ]; then
+  docker stop sigma-ldap
+  docker rm sigma-ldap
+  exit 0
+fi
+
+
+docker run -p 389:389 -p 636:636 --name sigma-ldap --env LDAP_TLS=false --env LDAP_ORGANISATION="sigmadev" --env LDAP_DOMAIN="sigma-dev.github.io"  --detach osixia/openldap:1.1.8
+sleep 3
```
adamschaub: Love it

auth/tokenizer/jwt.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/sigma-dev/sigma/auth/user"
+
+	"github.com/dgrijalva/jwt-go"
+)
+
+var (
+	ErrorNoID = errors.New("No ID in Token")
+)
+
+const DefaultTTL = int64(3600)
```
adamschaub: why not a `time.Duration`

auth/tokenizer/null.go
```diff
@@ -0,0 +1,26 @@
...
+func (j NullTokenizer) Tokenize(u user.User) (Token, error) {
+	return authToken{"null", 9999999}, nil
+}
+
+// Validate checks if token is legit.
+func (j NullTokenizer) Validate(token string) (bool, error) {
+	return true, nil
+}
+
+// ValidateAndGetID returns both a validity check as well as
```
adamschaub: `s/ValidateAndGetID/ValidateAndGet/`

auth/user/user.go
```diff
@@ -0,0 +1,86 @@
...
+	return int(user.Role())&int(role) >= int(role)
+}
+
+type User interface {
+	Name() string
+	Group() string
+	Role() Role
+}
+
+// TODO: ROLE/PERMISSION STUFF
```
adamschaub: Can we mark down what Role/Permission stuff entails? Either in Rally, an issue, whatever

auth/user/user.go
```diff
@@ -0,0 +1,86 @@
...
+	ROLE_RO
+	ROLE_ADMIN
+	ROLE_USERADMIN
+)
+
+func HasRole(user User, role Role) bool {
+	if role == ROLE_NIL {
+		return true
+	}
+	return int(user.Role())&int(role) >= int(role)
```
adamschaub: I like it

cmd/sigma.go
```diff
@@ -118,11 +122,28 @@ func sigmaRun(cmd *cobra.Command, args []string) {
 
 	// Sigma API Server
 	var svr *server.Server
-	if svr, globalErr = server.New(gCfg.Server, db); globalErr != nil {
+	//TODO: Once we have more auth methods, this'll change.
```
adamschaub: What's you mean

cmd/sigma.go
```diff
@@ -118,11 +122,28 @@ func sigmaRun(cmd *cobra.Command, args []string) {
...
+		// TODO: Config SECURED Shared Secret for JWT. (Maybe store in DB???)
+		tokener = tokenizer.NewJWT(gCfg.Auth.SharedSecret)
+	}
+	if method == "db" || method == "database" {
+		auther = authorizer.NewDB(db)
+	} else if method == "ldap" {
+		if auther, globalErr = authorizer.NewLDAP(gCfg.Auth.LDAP); globalErr != nil {
+			return
+		}
+	}
```
adamschaub: Constants for these strings

cmd/sigma.go
mikeynap: old todo, will delete

cmd/sigma.go
adamschaub: Can this initialization not be handled inside the auth package?

mikeynap: done.

database/couchbase.go
```diff
@@ -771,6 +767,15 @@ func (cb Couchbase) makeQuery(obj model.Metadataer, idOnly bool) (string, map[st
 		}
 	}
 
+	for k, v := range filters {
+		k = strings.Title(k)
+		if isClean(k) && isClean(v) {
+			query, isFirst = queryHelper(isFirst, query, fmt.Sprintf("%s=$%s", k, strings.ToLower(k)))
+			params[strings.ToLower(k)] = v
+		} else {
+			log.Warnf("Non-sanitary key value for query filter %v", k)
```
adamschaub: Fail the query if this happens

mikeynap: We don't fail for unsanitary metadata values, we just continue...

auth/authorizer/db.go
adamschaub: 19126058_10212712284857583_1548520186_o

auth/authorizer/db_test.go
adamschaub: also, we should add some captcha

auth/authorizer/ldap.go
adamschaub: So `domain` isn't `a.b.com`?

auth/authorizer/db_test.go
mikeynap: http://imgur.com/2u4knjR

auth/authorizer/ldap.go
mikeynap: it's either.

auth/authorizer/ldap.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env sh
```
adamschaub: Let's move this to integration. I don't see a reason for this to be in with the authorizer code.

adamschaub:
Few points then I'm GTM:

- Return message "Unauthorized" when an endpoint cannot be used. Status is coming through, but no response makes it initially confusing why a curl fails.
- Move the ldap init script to integration (currently in auth/authorizer)
- Add a configureable TTL for JWTs
- Stop logging `INFO[2017-06-14T01:18:29-04:00] Created default user "admin" with password "admin".  file=db.go line=74`, you donut
- When using ldap, "admin" is not returned when getting all users. When using db, it is. PICK ONE

auth/authorizer/ldap.go
adamschaub: Then your description is wrong wrong wrong

database/couchbase.go
adamschaub: Yeah well, I don't necessarily agree with that decision

auth/authorizer/ldap.sh
mikeynap: kk

mikeynap:
aight I implemented ^^^^. However, except for "Stop logging INFO[2017-06-14T01:18:29-04:00] Created default user "admin" with password "admin". file=db.go line=74"

Without this, no one can use anything. It's pretty common on first launch to give the default username/password isn't it?

mikeynap:
Talked with Matt -- he said not to print the password part, but to display the username on first login.

adamschaub:
LEGIT

Issue #133, opened 2017-06-08
Log aggregation for Sigma 


meganokeefe:
Service logs are currently stored in `Shipper` structs along with `Progress` and `Outputs`. But given the size of logs for long-running Services, this isn't sustainable in production.

We should explore the idea of piping/aggregating Sigma Service logs into a system meant for logs. (ie. an EFK stack or something similar). 

Whether this means we'll directly POST logs to an aggregation system, or have Sigma's API `UpdateServiceOperationLogs()` handle it is TBD.

PR #138, opened 2017-06-08
Tasker overhaul

### LINKED ISSUE #129: NomadTasker has a database dependency
Tasker should not rely on the database. Move all Tasker logic that requires the database (logs, progress, outputs -> the "Shipper") into the Manager.

meganokeefe:
Closes #129. 

## Changelog 
- [x] Removed `Tasker.Logs()`, `Tasker.Progress()`, and `Tasker.Outputs()`. Instead, now NomadTasker and DockerTasker both "ship" all three to the Sigma API --> DB. This removes NomadTasker's database dependency and standardizes the two Taskers.
- [x] Removed the `Shipper` struct and instead have isolated `ServiceOperation.Logs` into their own struct. `Progress` and  the raw string`Outputs` now live inside a `ServiceOperation`. 
- [x] ADD and GET API endpoints/functions for `Progress`, `Logs`, `Outputs`. All 3 have the url structure: `v1/services/<service-id>/op/<service-op-id>/logs`. (replace `logs` with `progress` or `outputs`). 
- [x] Updated database, model to reflect these changes 
- [x] Instead of calling `Tasker.Progress()`, etc., Manager now calls `Ship()` to get stuff into the DB so that on its next loop, ServiceOperations/ServiceOperationLogs have logs, progress, output.

- [x] refactored the server API tests 
- [x] updated Tasker tests (unit and integration) 
- [x] updated Manager tests (unit and integration) 

## TODO 

Logs will eventually become a problem here, given their size. Eventually, we should consider piping them into some log aggregation system VS. storing all serviceOp logs in couchbase. Put out an issue for this. #133


api/v1/server/routers.go
```diff
@@ -24,32 +24,23 @@ type Route struct {
 }
 
 func newRouter() *Router {
-	return &Router{gmux: mux.NewRouter().StrictSlash(false)}
-
+	return &Router{mux.NewRouter().StrictSlash(false)}
```
mikeynap: idk why your branch is getting precedence over master here... make sure you either merge master or ensure that the gmux stuff (line 13, 27, 259, (r *Router) ServeHTTP) are taken from master. (See https://github.com/sigma-dev/sigma/commit/e9f1975faa9b69a30ab01d4b2410667db1351127).

api/v1/server/server.go
```diff
@@ -238,65 +238,177 @@ func (s *Server) AddService(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, "Decoder error for API-derived serviceOp logs", http.StatusBadRequest)
+		return
 	}
-	w.Header().Add("Location", fmt.Sprintf("/v1/services/%v/shipper", theID))
+	if _, ok = apiObj["outputs"]; !ok {
+		log.Error("No outputs found in AddServiceOperationOutputs request")
+		http.Error(w, "No outputs found in AddServiceOperationOutputs request", http.StatusBadRequest)
+		return
+	}
+	parsedOutputs, err := parse(apiObj["outputs"].(string))
```
mikeynap: I think that this could panic easily (since Decoding to a map[string]interface{}). Have parse() take an interface{} and try to convert to a string in that function, as opposed to here ^.

Or just do it here... point is, you should have an "ok" check that this is actually a string.

mikeynap: Apply this to everywhere in the PR you see .(string) on decoded raw JSON

mikeynap:
It sorta is unfortunate that this is gonna use the raw API, since now we have to worry about authentication. 

We'll have to talk about what is the best way to address this -- both nomad && docker will both need to get JWT tokens to post to the API. We should have some sorta abstraction in place for that.

mikeynap:
This output appears a ton in the logs: 
```INFO[2017-06-12T18:00:11-04:00] Updated serviceOperation output for 30f43bb8-12d4-44e5-aca8-9f65d410a9cb-operation  file=server.go line=411
INFO[2017-06-12T18:00:11-04:00] Updated ServiceOperation progress: 4de93ea1-c8ad-41a4-a7d4-059e75a9bd6d-operation  file=server.go line=357
WARN[2017-06-12T18:00:14-04:00] Empty outputs for serviceOperation: 4de93ea1-c8ad-41a4-a7d4-059e75a9bd6d-operation  file=manager.go line=443
WARN[2017-06-12T18:00:14-04:00] About to delete service operation: 4de93ea1-c8ad-41a4-a7d4-059e75a9bd6d-operation  file=manager.go line=500
```

I think some or most of these can be changed to debug

api/v1/server/routers.go
meganokeefe: Fixed

manager/manager.go
```diff
@@ -417,12 +417,14 @@ func (m *manager) statusOperation(op model.ServiceOperation) {
...
 		return
 	}
+	// Progress already in Op - no need to update
 	log.Debugf("%s %s operation: Progress- %s, Status-%s", op.Service.Name, op.Operation, op.Progress, op.Status)
 }
 
 // finishOperation, called when an operation succesfully completes
 func (m *manager) finishOperation(op model.ServiceOperation) (retErr error) {
 	defer m.updateOperation(&op)
+	log.Warnf("In FinishOperation, operation = %s", op.Operation)
```
mikeynap: ?

api/v1/server/server.go
mikeynap: Or just change to this: 
```
	output, ok := apiObj["outputs"]
	outputStr, ok2 := output.(string)
	if !ok || !ok2 {
		log.Error("No outputs found in AddServiceOperationOutputs request")
		http.Error(w, "No outputs found in AddServiceOperationOutputs request", http.StatusBadRequest)
		return
	}

	parsedOutputs, err := parse(outputStr)
```

api/v1/convert/convert_test.go
```diff
@@ -41,26 +41,26 @@ func TestSite(t *testing.T) {
 	}
 }
 
-func TestShipper(t *testing.T) {
-	shipper := struct {
-		m model.Shipper
-		a apiclient.Shipper
+func TestServiceOperationLogs(t *testing.T) {
```
mikeynap: Quality Code. Good Test. Very Worth.

adamschaub: +1 agreed, what would we do without this test?

api/v1/server/server.go
```diff
@@ -251,64 +251,192 @@ func (s *Server) AddService(w http.ResponseWriter, r *http.Request) {
...
+	if len(foundLogs) == 0 {
+		if _, err := s.db.InsertServiceOperationLogs(modObj); err != nil {
+			log.Errorf("ServiceOperationLogs failed insert: %v", err)
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
-		log.Infof("Added Shipper: %s", modObj.Name)
-	} else { // Update shipper if exists
-		toUpdate, err := s.db.GetShipper(foundShipper[0])
+		log.Infof("Added ServiceOperationLogs for: %s", serviceOpId)
```
adamschaub: Make debug

adamschaub: Or just remove. We can see that logs have been added by looking at the logs

manager/manager.go
```diff
@@ -371,6 +370,8 @@ func (m *manager) updateOperations(op model.ServiceOperation) {
 		if err := m.finishOperation(op); err != nil {
 			log.WithError(err).Warnf("Unable to finish operation %s. Skipping.", op.Id)
 			return
+		} else {
```
adamschaub: Get rid of the else, just log after the if closes

manager/manager.go
```diff
@@ -500,14 +491,17 @@ func (m *manager) removeOperation(op model.ServiceOperation) {
 	}
 }
 
-func (m *manager) updateOperation(op *model.ServiceOperation) {
-	// Save changes to service operation
-	if op == nil {
-		return
+// updates service operation status
+func (m *manager) updateStatus(opId string, stat string) {
```
adamschaub: let's call it `updateOperationStatus` or something a little more specific. Services have status too.

task/docker.go
```diff
@@ -237,21 +217,103 @@ func (d *DockerTasker) Status(so *model.ServiceOperation) (string, error) {
...
+	output := ""
+	rawOutput, err := ioutil.ReadFile(path.Join(d.saveDir, so.Service.Id, "output"))
+	if err == nil {
+		output = base64.StdEncoding.EncodeToString([]byte(rawOutput))
+	}
+	return logs, progress, output
+}
+
+// postShipper is a helper for Ship()-  POSTs logs/progress/output to the sigma API
+func (d *DockerTasker) postShipper(so *model.ServiceOperation, payload []byte, endpointSuffix string) error {
```
adamschaub: Why aren't we using the API client to do this call?

task/docker_test.go
```diff
@@ -34,10 +38,18 @@ func testServiceOperation(image string, driverConfig map[string]interface{}) *mo
...
+	cfg       DockerConfig
+	useAPIUrl = "127.0.0.1:24609"
+)
 
 func TestMain(m *testing.M) {
 	cfg.SaveDir = "./_docker_temp_test"
+	cfg.APIUrl = "127.0.0.1:24609"
+
+	db := database.NewNullDB(false)
+	go startAPI(db)
```
adamschaub: We need to avoid this kind of stuff in unit tests.

task/docker_test.go
```diff
@@ -488,3 +484,18 @@ func TestDockerPull(t *testing.T) {
 		}
 	}
 }
+
+func startAPI(db database.Database) {
```
adamschaub: This isn't kosher for unit tests.

tests/integration/manager/manager_test.go
```diff
@@ -67,16 +69,18 @@ func TestManager(t *testing.T) {
 func ActuallyEqual(i interface{}, val []string) bool {
```
adamschaub: This is broken. I didn't catch this with my types update. If manager tests are failing, this is most likely the culprit.

api/v1/client/service_operation_outputs.go
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Generated by: https://github.com/swagger-api/swagger-codegen.git
+ */
+
+package apiclient
+
+type ServiceOperationOutputs struct {
+	Id string `json:"id,omitempty"` //serviceOperation ID
+
+	Outputs string `json:"outputs,omitempty"`
```
meganokeefe: Nomad reasons and maybe some of my own laziness. 

We do the `curl POST` for nomad job outputs from an inline script that's passed around like so: https://github.com/sigma-dev/sigma/blob/taskerv2/task/nomad.go#L218

And ships outputs as a raw string, and then manager parses them out of the database.  

I could figure out how to unmarshal the outputs as a json within the script, but it was basically about not wanting to make this terrible script any longer.

task/docker.go
meganokeefe: that, my friend, is a very good question.

tests/integration/manager/manager_test.go
meganokeefe: @mikeynap can you explain your changes to this helper? thx

PR #141, opened 2017-06-12
Fix and tests for duplicate names for Plugin, Service, and Bundle

### LINKED ISSUE #107: Disallow posting services, other objects with same name twice through API
Can post same service multiple times, each document gets own ID.

Need to guarantee service uniqueness based on Name & Site.

dlb46:
Closes #107 

### Change log

Added functionality in server.go to disable duplicate names for Plugins, Services, and Bundles.
Also added tests in server_test.go to make sure it works.

### Notes

Everything seems to be working.

### TODO

I did this for Plugins, Services, and Bundles.  I think this is all that need it, but there may be others that still need name validation that I do not know about.


api/v1/server/server.go
```diff
@@ -890,6 +907,49 @@ func (s *Server) validateSite(site apiclient.Site, isUpdate bool) error {
 	return nil
 }
 
+//validatePlugin is a helper that validates a Plugin before Add and Update
+func (s *Server) validatePlugin(plugin apiclient.Plugin, isUpdate bool) error {
+	//1. Plugin cannot have duplicate names
+	if !isUpdate {
```
meganokeefe: don't we want to validate in the case where `if isUpdate { ` ? aka isUpdate=`true`?

dlb46: 1) I used the ValidateSite function as a template, and they do !isUpdate
2) I think this makes sense because if we check for duplicate names when the Plugin, Server, or Bundle is created, and disable the ability to change the name once it is created (isUpdate = false), then no duplicate names can happen

meganokeefe: My bad -- you are correct, we want to check duplicates on new object adds, updates don't matter.

meganokeefe:
LGTM!

PR #142, opened 2017-06-12
Delete has priority over Status on active service

stephenrlouie:
### Change log

- Delete has the highest priority when a service is active
- PR template update (should be another PR sorry)


manager/manager.go
```diff
@@ -313,18 +313,18 @@ func (m *manager) updateService(serviceID string) {
 		}
 
 	case model.ServiceStatusActive:
-		if inputsChanged {
+		if !service.IsEnabled() || model.BoolValue(service.Remove) {
```
vgunning: Do we want if service isn't enabled AND we want to remove the service?
I think we may end up deleting services that just currently aren't enabled here...

adamschaub: It doesn't delete, the service, just stops it. A service shouldn't be active unless it was enabled. If it is active and not enabled, we stop the service.

PR #144, opened 2017-06-12
Choose nomadNodeID from attributes

meganokeefe:
Right now we're not constraining on "hints" for NomadTasker -- this fixes that.


vgunning:
LGTM

PR #147, opened 2017-06-13
Fix and test for Bug #139, Plugin.Name should be immutable

### LINKED ISSUE #139: Plugin.Name should be immutable
reject updatePlugin if the plugin has a name change.

dlb46:
Fix #139 

### Change log

Once a plugin is created, update should disabled if they change the name.  To accomplish this I compared the old name to the new name, and if they do not match then return a StatusBadRequest (400) error

### Notes

Only question is if a StatusBadRequest is the correct error to return

### TODO

Don't think so


api/v1/server/server.go
```diff
@@ -741,6 +741,10 @@ func (s *Server) UpdatePlugin(w http.ResponseWriter, r *http.Request) {
 		updatedPlugin := convert.FromAPIPlugin(apiObj)
 		updatedPlugin.MetaImpl = plugin2Update.MetaImpl
 		updatedPlugin.Id = plugin2Update.Id
+		//plugin2Update is the old plugin and updatedPlugin is the new
+		if plugin2Update.Name != updatedPlugin.Name {
+			return http.StatusBadRequest, fmt.Errorf("Cannot change name of plugin")
```
meganokeefe: Small change -- could you create an error `var` (like the ones here: https://github.com/sigma-dev/sigma/blob/master/api/v1/server/server.go#L31) that contains the string "Cannot change name of plugin"?  And then do a `return http.StatusBadRequest, fmt.Errorf(ErrPluginNameChanged` or something? 

This lets us pass around errors in a clearer way -- not super meaningful right now because we're only calling this error once so far, but more of a best-practices thing?

PR #150, opened 2017-06-13
Strip cas

vgunning:
### PR checklist

### Change log

- Added a MetaImpl.Strip function to remove a key from the Metadata map.
- Updated Insert and Update Couchbase functions to strip any CAS value in the object's metadata.
- Consolidated all string type declarations to exported constants example: `ServiceType = "SERVICE"`
- Standardized all Update and Insert Couchbase functions to add the object type to the metadata.


model/model.go
```diff
@@ -59,6 +59,17 @@ func (m *MetaImpl) Get(k string) (interface{}, bool) {
 	return v, ok
 }
 
+//Strip removes the key from the map if it exists
+func (m *MetaImpl) Strip(k string) {
+	if m.Metadata == nil {
+		m.Metadata = make(map[string]interface{})
+	}
+	_, ok := m.Metadata[k]
```
mikeynap: This is unneeded. no need to check if exists before calling delete.

mikeynap:
Strip The Casbah

database/couchbase.go
```diff
@@ -385,26 +383,35 @@ func (cb *Couchbase) InsertSite(m model.Site) (string, error) {
 
 func (cb *Couchbase) UpdatePlugin(m model.Plugin) error {
 	m.Set("Type", PluginType)
+	m.Strip("cas")
```
adamschaub: Stripping cas in any of the "update" functions before the update happens will cause nothing to pass couchbase's cas check. Move `m.Delete("cas")` into `update` after we've fetched the cas from metadata.

database/couchbase.go
```diff
@@ -364,6 +360,7 @@ func (cb *Couchbase) InsertBundle(m model.Bundle) (string, error) {
 
 func (cb *Couchbase) InsertServiceOperation(m model.ServiceOperation) (string, error) {
 	m.Set("Type", ServiceOperationType)
+	m.Strip("cas")
```
adamschaub: If this is true for every update, modify the `insert` func to do the stripping. You'll need to pass in the model as a Metadater interface.

model/model.go
```diff
@@ -59,6 +59,17 @@ func (m *MetaImpl) Get(k string) (interface{}, bool) {
 	return v, ok
 }
 
+//Strip removes the key from the map if it exists
+func (m *MetaImpl) Strip(k string) {
```
adamschaub: Unnecessary, just use `Delete`

PR #151, opened 2017-06-13
Fix and test bug 140, reject plugin delete if existing services use it

### LINKED ISSUE #140: Reject Plugin delete if existing services use it
A plugin can only be deleted if there are no services that depend on that plugin.

Relevant place for bug fix: https://github.com/sigma-dev/sigma/blob/master/api/v1/server/server.go#L355

dlb46:
Fix #140 

### Change log

In DeletePlugin, get all services, loop through their plugins, and see if they match the plugin they are trying to delete

To test, just try to delete a plugin that is being used by a service

### Notes

Is the error message "Cannot delete a plugin being used by a service" good?

### TODO

Don't think so


meganokeefe:
LGTM!

PR #154, opened 2017-06-14
Dakota/bug91

### LINKED ISSUE #91: Enabled should be set to true by default
I feel like the cases where one would want a service to be enabled upon adding is more common than the current "enabled: false" default. 

Adding a service already depends on other services being ready, I think it makes sense to have them enabled unless specifically told NOT to be enabled. 

Just a thought.

(if we decide this is stupid and I should shut up, maybe a /v1/servies/{Sid}/enable would be a convenient endpoint).

dlb46:
Fix #91 

### Change log

Added a constructor function in service.go called NewServiceConfig().  This is what sets the Enabled to true by default.  Also, went through the code and changed what used to be apiclient.Service to apiclient.NewServiceConfig() at all the locations I thought needed it.  Finally, I added a test called TestDefaultEnabled in server_test.go.

### Notes

Now, in order to get a service with the Enabled defaulted to true, you must call NewServiceConfig().  Notice this function has no input, so you must then populate the Service after you call NewServiceConfig.  An example:

var testService = apiclient.Service{Name: "test"}

is now,

var testService = apiclient.NewServiceConfig()
testService.Name = "test"

### TODO

If someone else could comb through the code and look for other places where this new constructor function should be called, instead of the old (not defaulted) way, that would be great.


api/v1/client/service.go
```diff
@@ -38,3 +38,9 @@ type Service struct {
 
 	Enabled bool `json:"enabled,omitempty"`
 }
+
+func NewServiceConfig() Service {
+	return Service{
+		Enabled: true,
+	}
+}
```
adamschaub: This file is auto-generated by swagger (see https://github.com/sigma-dev/sigma/tree/master/api/v1/gen), so we need to avoid modifying the file directly.

If you modify swagger's `v1.yaml` definition, you might be able to regenerate the service with `enabled` defaulting to true: http://swagger.io/unlocking-the-spec-the-default-keyword/

adamschaub:
Looks good!

PR #156, opened 2017-06-14
Ensures Integration Tests Can Fail

stephenrlouie:
## Change log

- Ensures the make integration will cause the build to fail, was passing no matter what before (have a new make directive, `make clean_integration`
- Repaired nomad_tests
- Attempts to fix manager_tests, but seems to be a dependency on the types system and other problems that I wanted to merge this in so others can benefit now. (need a few days to figure out the types and manager work)

TODO:
- #157 : @mikeynap and/or @meganokeefe is on it.


docker_push.sh
```diff
@@ -1,16 +1,21 @@
...
-    docker tag $SIGMA_REPO:$SIGMA_TAG $PROD_REPO:$TRAVIS_TAG
-    docker push $PROD_REPO
+  if [ ! -z "$TRAVIS_TAG" ]; then
+      make container
+      docker login -e $DOCKER_EMAIL -u $DOCKER_USER -p $DOCKER_PASS
+      docker tag $SIGMA_REPO:$SIGMA_TAG $PROD_REPO:$TRAVIS_TAG
+      docker push $PROD_REPO
+  fi
+else
+  echo "Build is broken"
```
meganokeefe: maybe this is my lack of bash knowledge -- but to which `if` does that `else` correspond? wouldn't we want to say "Build is broken" if `TRAVIS_TEST_RESULT != 0`?

meganokeefe:
Changes LGTM-- As you note, manager tests still broken though.
Also have you tried doing `[integration]` to test the `clean` stuff in travis?

docker_push.sh
stephenrlouie: thats the exit code of the build, so if the build is known to have failed we will not push the container. Before we would push no matter what

stephenrlouie:
@meganokeefe if you look at the travis build, its already always running `make integration`

PR #158, opened 2017-06-15
Change CreateLock interface. Wrap body in metadater before insert

adamschaub:
### Change log

Wrap CreateLock body in a Metadataer


database/mock_database/mock_database.go
```diff
@@ -4,10 +4,9 @@
 package mock_database
 
 import (
```
vgunning: NIT?

PR #159, opened 2017-06-16
Safe Site Deletion

stephenrlouie:
### Change log

~- Will reject site deletions if services reference it.~

Updated: 7.5.17 1PM

- Mark for deletion flag added to Sites
- Garbage collection will occur every so often(1m), looking at the sites marked for deletion and ensure there are no references on them. Printing when references are found and when a deletion is successful
- Rejects Sites and Services referencing a site that is marked for deletion
- Exposed the deletion flag through the api

### TODO

- [x] block deleting the default site
- [x] this needs tests. 
- [x] ~default running site if we cannot find the desired site.~ Not an issue because we cannot delete sites if services reference them.
- [x] MARK FOR DELETION : Sites will be marked for deletion
   - [x] needs to block these sites from getting new site children
   - [x] needs to block services from using this site
   - [x] exposes the field in swagger
   - [x] ensure DB calls will properly handle Remove
   - [x] ensure converts will handle Remove
   - [x] Test queries on marked for delete sites

~@mikeynap some of the convert values changed on User Token. Can you look at the changes I made due to the re-generation of the client via swagger.~


api/v1/server/server.go
```diff
@@ -1220,6 +1222,17 @@ func (s *Server) UpdateService(w http.ResponseWriter, r *http.Request) {
 		updatedService.MetaImpl = service2Update.MetaImpl
 		updatedService.Status = service2Update.Status
 		updatedService.Id = service2Update.Id
+
+		doc, err := s.db.GetSite(updatedService.Site)
+		if err != nil {
+			return http.StatusNotFound, fmt.Errorf("Could not get the services Site: %v", err)
```
vgunning: Also yay.

api/v1/server/server_test.go
```diff
@@ -181,6 +181,7 @@ func TestValidateService(t *testing.T) {
 	scrap.Name = "scrap"
 	scrap.Plugin = emptyPlugin.Name
 	scrap.PluginVersion = emptyPlugin.Version
+	scrap.Site = "local"
```
vgunning: `model.DefaultSite`?

api/v1/server/server_test.go
```diff
@@ -207,19 +208,18 @@ func TestValidateService(t *testing.T) {
...
 	}{
-		{scrap, fmt.Errorf(ErrDuplicateName)},
-		{noName, fmt.Errorf(ErrInvalidName)},
-		{badName, fmt.Errorf(ErrInvalidName)},
-		{unknownSite, fmt.Errorf(ErrUnknownSite)},
-		{noSite, fmt.Errorf(ErrUnknownSite)},
+		{scrap, ErrDuplicateName},
+		{noName, ErrInvalidName},
+		{badName, ErrInvalidName},
+		{unknownSite, nil},
```
vgunning: These default to the `DefaultSite` now?

api/v1/server/server_test.go
```diff
@@ -490,17 +494,106 @@ func TestDeleteService(t *testing.T) {
...
+		postSiteToAdd    apiclient.Site
+		siteErrCode      int
+		postServiceToAdd apiclient.Service
+		serviceErrCode   int
+	}{
+		{ // delete no parent
+			sitesToAdd:      []apiclient.Site{massSite, bxbSite, bosSite},
+			servicesToAdd:   []apiclient.Service{},
+			siteToDelete:    bosSite.Name,
+			expectedErrCode: 200,
```
vgunning: Can we put in the `http` status codes from the library instead like `http.StatusOK`? maybe a bit more readable.

database/mock_database/mock_database.go
```diff
@@ -1,14 +1,13 @@
...
 
 package mock_database
 
 import (
-	time "time"
-
 	gomock "github.com/golang/mock/gomock"
 	database "github.com/sigma-dev/sigma/database"
 	model "github.com/sigma-dev/sigma/model"
+	time "time"
```
vgunning: This is rather amusing.

database/nulldb.go
```diff
@@ -422,7 +422,7 @@ func (n *NullDB) QuerySites(o model.Site) ([]model.Site, error) {
 			mdMatch = mdMatch && mdv == v.Metadata[k]
 		}
 
-		if mdMatch && (o.Id == "" || o.Id == v.Id) && (o.Parent == "" || o.Parent == v.Parent) && (o.Name == "" || o.Name == v.Name) {
+		if mdMatch && (o.Id == "" || o.Id == v.Id) && (o.Name == "" || o.Name == v.Name) && (o.Parent == "" || o.Parent == v.Parent) && (o.Remove == nil || model.BoolValue(o.Remove) == model.BoolValue(v.Remove)) {
```
vgunning: Would it be better to make this huge if statement a tiny helper function?

manager/manager.go
```diff
@@ -161,6 +172,10 @@ func (m *manager) Start() error {
 			m.workerWg.Wait()
 			log.Info("Quitting Sigma Manager")
 			return nil
+		case <-gcTicker.C:
+			// garbage collect Sites
+			log.Debug("Garbage collecting sites")
+			m.cleanSites()
 		case <-ticker.C:
```
vgunning: NIT: Rename ticker to its purpose? `svcTicker`?

manager/manager.go
```diff
@@ -161,6 +172,10 @@ func (m *manager) Start() error {
 			m.workerWg.Wait()
 			log.Info("Quitting Sigma Manager")
 			return nil
+		case <-gcTicker.C:
+			// garbage collect Sites
+			log.Debug("Garbage collecting sites")
+			m.cleanSites()
```
vgunning: Do we want to log errors that are returned by the `m.cleanSites()`? or is it for testing purposes only

manager/manager.go
```diff
@@ -205,6 +219,59 @@ func (m *manager) Stop() {
...
+		return err
+	}
+	for _, id := range siteIDList {
+		//check if I have children
+		myself, err := m.db.GetSite(id)
+		if err != nil {
+			return err
+		}
+
+		if myself.Name != model.DefaultSite {
```
vgunning: simplify logic/nesting by changing to 
```
if myself.Name == model.DefaultSite {
   continue
```

manager/manager.go
```diff
@@ -205,6 +219,59 @@ func (m *manager) Stop() {
...
+// Looks for sites marked for deletion, deletes ones that are clear of children
+func (m *manager) cleanSites() error {
+	siteIDList, err := m.db.QuerySitesIds(model.Site{Remove: model.NewBool(true)})
+	if err != nil {
+		return err
+	}
+	for _, id := range siteIDList {
+		//check if I have children
+		myself, err := m.db.GetSite(id)
+		if err != nil {
```
vgunning: Do we want to completely error out if we fail to get one site by its `id`?

manager/manager_test.go
```diff
@@ -256,6 +256,121 @@ func expectStartOperation(t *testing.T, m *manager, db *mock_database.MockDataba
 	tskr.EXPECT().Run(&op).Return(nil)
 }
 
+// a nil valie indicates that the corresponding call will not be executed by cleanSites
+func TestCleanSites(t *testing.T) {
```
vgunning: Have something with a mix of references by `id` and `name`?

api/v1/server/server_test.go
stephenrlouie: yes

api/v1/server/server_test.go
stephenrlouie: truth

manager/manager.go
stephenrlouie: will log

manager/manager.go
stephenrlouie: nah we will continue to try the others

manager/manager_test.go
stephenrlouie: yup

database/nulldb.go
stephenrlouie: @mikeynap's PR addresses: #196

PR #161, opened 2017-06-19
Log aggregation for internal logs + plugin logs

### LINKED ISSUE #133: Log aggregation for Sigma

meganokeefe:
Closes #133 and #169 

Added an EFK pipeline for log streaming -- support for sigma's own logs, plus plugin logs. `deploy.sh` will start up containers for ElasticSearch, Fluentd, and Kibana. Note that fluentd must run on `0.0.0.0` so that remote (nomad) hosts can forward their logs back home.

The Kibana UI can be reached at `http://127.0.0.1:5601/.` Note you'll have to set an `index pattern` before you can view/visualize the logs. The index pattern is `fluentd-<YYYYMMDD>`, with today's date.

## Changelog 

- [x] deploy + stop scripts for the EFK stack. right now, using docker-compose for this. (see TODO)
- [x] sending sigma internal logs to EFK, via the logrus/fluentd hook (vendored). In kibana, you'll see: `@log_name:sigma-internal` to identify internal logs. 
- [x] For Docker tasker, setting plugin containers' `log-driver=fluentd`. Plugin logs are identifiable by `<service-name>:<service-operation-id>` (see TODO). 
- [x] For Nomad tasker, setting the configurable `logging` options to use fluentd. Because nomad clients are remote hosts, added a config field for the sigma host's IP, to use as `fluentd-address`.
- [x] Removed log posting from the shipper script and from `dockertasker.Ship()`, but keeping the HTTP endpoints for logs (see TODO).  
- [x] Updated `NomadTasker.Status()` to change the way we determine if a nomad job is "done." (With the Service scheduler, nomad jobs will never reach the Complete state that manager wants, since the container is deployed over and over by Nomad.) Using [Nomad allocation TaskStates](https://www.nomadproject.io/docs/http/alloc.html#taskstates), now, as status indicators. 

## TODO 
1. Configurable fluentd forwarding output plugins, for Sigma. Right now we're defaulting to ElasticSearch, but should also let users forward to the sigma API -> couchbase (#165). (and potentially other output destinations.)
2. Transition away from [docker-compose for EFK](http://docs.fluentd.org/v0.12/articles/docker-logging-efk-compose). Build a custom sigma EFK image and deploy to sigma-dev. Need to figure out how to do it though.
3. Look into fluentd log tagging / adding fields re: plugin logs -- right now, we're concatenating `<service-name>:<service-op-id>` for the fluentd tag, of which it seems you can only have one. Ideally we use something like [labels](https://docs.docker.com/engine/admin/logging/fluentd/#usage) (which I have yet to get working with the Docker or Nomad APIs).


helik:
should you be committing config.yaml?

tests/integration/stop.sh
```diff
@@ -9,5 +9,5 @@ echo "Killing couchbase..."
 echo "Killing rabbit..."
 ./queue/stop.sh
 
-echo "Killing LDAP"
+echo "Killing LDA...P"
```
helik: was this intended?

helik:
Cool, tested it out and it works so lgtm =)

task/nomad.go
```diff
@@ -79,61 +74,92 @@ func (nt *NomadTasker) Remove(so *model.ServiceOperation) error {
...
-		outStatus = model.ServiceOperationStatusWaiting
-	case "dead":
-		outStatus = model.ServiceOperationStatusFailed
-	case "complete":
-		outStatus = model.ServiceOperationStatusFinished
-	default:
-		outStatus = model.ServiceOperationStatusUnknown
-	}
-	return outStatus, nil
+	badStates := []string{"Setup Failure", "Driver Failure", "Killing", "Killed", "Failed Validation", "Not Restarting", "Sibling Task Failed", "Leader Task Dead"}
```
stephenrlouie: could we use the consts instead. https://github.com/hashicorp/nomad/blob/aac319cd169fa58139c6b7825855171cb97c2692/api/tasks.go#L456-L475

task/nomad.go
```diff
@@ -166,66 +179,51 @@ func (nt *NomadTasker) CreateNomadJob(so *model.ServiceOperation) (*nomadapi.Job
...
-	serviceTask := &nomadapi.Task{Name: "main", Leader: true, Driver: "docker", Env: env, Config: taskConfig, Templates: templates}
+	return &nomadapi.Task{Name: "plugin", Leader: true, Driver: "docker", Env: env, Config: taskConfig, Templates: templates}, nil
+}
 
-	// Shipper is an isolated exec task that can access main's logs, progress, and outputs
-	shipperScript := populateShipper(so.Service.Name, so.Id, nt.APIEndpoint)
-	shipperTask := &nomadapi.Task{
+// The shipper task handles the CURLing of Progress and Outputs to the Sigma API
+func (nt *NomadTasker) createShipperTask(so *model.ServiceOperation) (*nomadapi.Task, error) {
+	shipperScript := fmt.Sprintf(`openBracket={;
```
stephenrlouie: barf

task/nomad_test.go
```diff
@@ -90,28 +90,28 @@ func TestNomadStatus(t *testing.T) {
 		modelStat  string
 	}{
 		{
-			"running",
+			"Started",
```
stephenrlouie: consts?

stephenrlouie: are these the nomad consts?

task/docker.go
```diff
@@ -141,13 +141,13 @@ func (d *DockerTasker) Run(so *model.ServiceOperation) error {
 	}
 
 	hostConfig := &container.HostConfig{
-		Binds: []string{fmt.Sprintf("%s:%s", dir, OutputDirDefault)},
+		Binds:     []string{fmt.Sprintf("%s:%s", dir, OutputDirDefault)},
```
mikeynap: what if fluentd isn't setup?

task/nomad.go
```diff
@@ -20,46 +20,41 @@ var (
...
 
-// NewNomadTasker instantiates a new NomadTasker
+// NewNomadTasker instantiates a new NomadTasker by connecting to the nomad API client
 func NewNomadTasker(inConf NomadConfig) (*NomadTasker, error) {
-
-	//init random generator (used for generating the Nomad job / Node selection)
-	rand.Seed(time.Now().Unix())
-
-	// connect to the nomad API
+	rand.Seed(time.Now().Unix()) //random gen needed for chooseNomadNode (choose site)
```
mikeynap: move to `func init(){ }`

task/nomad.go
```diff
@@ -166,66 +188,51 @@ func (nt *NomadTasker) CreateNomadJob(so *model.ServiceOperation) (*nomadapi.Job
...
+		outputsUrl=http://%v/v1/services/%s/operation/%s/outputs;
+		echo $outputsUrl;
+		curl -X POST --data $outContents $outputsUrl;
+		`, nt.APIUrl, so.Service.Name, so.Id, nt.APIUrl, so.Service.Name, so.Id)
 
-	// Shipper is an isolated exec task that can access main's logs, progress, and outputs
-	shipperScript := populateShipper(so.Service.Name, so.Id, nt.APIEndpoint)
-	shipperTask := &nomadapi.Task{
+	cpu := 100
+	memory := 300
```
mikeynap: What're these ^^^? Should they be configurable/do they need to be provided? Where did you get these numbers from?

task/nomad.go
```diff
@@ -79,61 +74,101 @@ func (nt *NomadTasker) Remove(so *model.ServiceOperation) error {
...
-		outStatus = model.ServiceOperationStatusWaiting
-	case "dead":
-		outStatus = model.ServiceOperationStatusFailed
-	case "complete":
-		outStatus = model.ServiceOperationStatusFinished
-	default:
-		outStatus = model.ServiceOperationStatusUnknown
-	}
-	return outStatus, nil
+	badStates := []string{
```
mikeynap: add this as a var in the package namespace (top of file)

mikeynap:
```ERRO[2017-06-28T11:32:01-04:00] Could not start operation 2d8f8c76-6789-4e31-9bae-110f47f29e0f-operation  error="Error response from daemon: Failed to initialize logging driver: dial tcp 127.0.0.1:24224: getsockopt: connection refused" file=manager.go line=386```

This appears if nothing is specified in the config. There should be defaults that just work... right? Or do we REQUIRE this now?

task/nomad.go
meganokeefe: These were leftover from debugging, thought nomad was somehow running out of memory but it wasn't - removing!

task/docker.go
meganokeefe: Good catch, fixing this - fluentd isn't enabled by default in the tasker.

PR #164, opened 2017-06-20
Provide defaults for database and tasker configuration

### LINKED ISSUE #160: Run sigma with config
We should have defaults such that you shouldn't have to supply a config (or any command line flags) when running Sigma.

kochman:
### Change log

Added defaults so that Sigma can find Couchbase and the tasker can reach the API when Sigma is run without any configuration.

### Notes

Resolves #160.

One possible issue is that if Sigma is configured to listen on a non-default interface or port, then the tasker will fail with a TCP connection connection error if it is not also configured to connect to the new interface/port. This is expected, but the relationship between those configuration options may be confusing and require a note in the docs.


helik:
LGTM

PR #167, opened 2017-06-21
Make Functional Tests Great Again

stephenrlouie:
### Change log

- Turns off auth on functional tests
- Timeouts on the deploy, instead of indefinitely waiting


config.yaml
```diff
@@ -16,7 +16,7 @@ Couchbase:
   MBQuota: 100
   Timeout: 32s
 Auth:
-  Method: db #ldap,db,none
+  Method: none #ldap,db,none
   SharedSecret: stupidPants
```
meganokeefe: you should be able to remove the rest of the Auth stuff from the config, here.

database/deploy.sh
```diff
@@ -1,14 +1,21 @@
 ADDR="127.0.0.1"
 docker run -d --name sigma-db -p 8091-8094:8091-8094 -p 11210:11210 couchbase
-
-until $($CMD curl --output /dev/null --silent --head --fail http://$ADDR:8091); do
+COUNT=0
+LIMIT=15
+until $($CMD curl --output /dev/null --silent --head --fail http://$ADDR:8091) || [ $COUNT -eq $LIMIT ]; do
   printf '.'
   sleep 2
+  ((COUNT++))
```
meganokeefe: why the double-parentheses around this?

stephenrlouie: bash reasons

PR #168, opened 2017-06-21
Reduce manager's default intervals

### LINKED ISSUE #115: Adjust default time intervals to more reasonable values
My initial defaults for event loop updates are way too long. Need to play around with the values a bit to make Sigma more responsive out-of-the-box (but still stable). 

```
	DefaultMinOperationStatusInterval = "10s"
	DefaultStatusInterval             = "20s"
	DefaultUpdateInterval             = "1s"
	DefaultRetryInterval              = "20s"

	DefaultServiceInitInterval = "10s"
```

kochman:
### Change log

Decreased default update intervals in the manager event loop.

### Notes

Resolves #115.

I doubled DefaultMaxRetries because of the quartering of DefaultRetryInterval.

I noticed no slowdowns with these defaults, but wasn't terribly scientific about it.

For DefaultServiceInitInterval, I noticed anything lower than about four or five seconds could result in the queue bloating, so I didn't reduce it by much. This could be shorter on faster hardware, but it's just a default.


helik:
lgtm lgtm

PR #171, opened 2017-06-21
First fixes for bug #152

### LINKED ISSUE #152: /services does not return a service status
We need an API for getting a summary of all services for the purposes of displaying in a GUI table. Considering adding a /services?view=summary endpoint that provides a summary of each service, including its status.

dlb46:
Fix #152 

### Change log

For whatever reason, in convert.go we simply weren't passing the state of the service.  I changed it so that now we do.

In v1.yaml I added this "state" field to the Service api object.  This allows the state property to be passed back and forth.

In server.go I just made a constant for that string per convention.

In server_test.go I just made a service with a status, and checked to make sure it was being added and retrieved correctly.

### Notes

I talked to Craig Micallef, and he said that all he needed was the status of the service, so that is why I didn't add in the rest of the serverstatus stuff.  I told him if down the road he decides he wants ends up wanting that information (lastUpdated, logs, description, output), then I will add it there for him.  Doing that should be very intuitive. 

Make sure to run the Makefile in the /gen directory, and replace the existing "clients" folder with the one generated by swagger.

### TODO

I believe the online "Sigma API Specification" page needs to be changed so that this state attribute is added to the GET /Services.  I'm happy to do this, just ping me with the best way to do it.


api/v1/gen/v1.yaml
```diff
@@ -655,6 +654,12 @@ definitions:
           $ref: '#/definitions/ServiceIO'
       enabled:
         type: boolean
+      state:
+        type: string
+        enum:
```
meganokeefe: reason for this `enum` here vs. just a `string`? 
ServiceOperation status can be more things than this, also: https://github.com/sigma-dev/sigma/blob/master/model/model.go#L220

api/v1/server/server_test.go
```diff
@@ -605,6 +605,9 @@ func TestGetServices(t *testing.T) {
 	if ss, _, _ = client.GetServices("fullPlugin"); len(ss) != 1 {
 		t.Errorf("GetServices with Plugin-Name=fullPlugin should have returned 1 result. Returned %v instead", len(ss))
 	}
+	if ss, _, _ = client.GetServices("serviceWithStatusPlugin"); len(ss) != 1 || ss[0].State != serviceWithStatus.State {
+		t.Errorf("GetServices with Plugin-Name=servieWithStatus should have returned 1 result. Returned %v instead", len(ss))
```
meganokeefe: small typo, `servie` -> `service`

api/v1/convert/convert.go
```diff
@@ -76,6 +76,7 @@ func FromModelService(o model.Service) apiclient.Service {
 		Plugin:        o.PluginName,
 		Savedir:       o.SaveDir,
 		PluginVersion: o.PluginVersion,
+		State:         o.Status,
```
meganokeefe: small thing, can you change `State` to `Status`, no real reason to rename this field, yes?

api/v1/gen/v1.yaml
```diff
@@ -655,6 +654,12 @@ definitions:
...
+          - "stopped"
+          - "inactive"
+          - "running"
 
   ServiceStatus:
```
meganokeefe: I'm inclined to say that we can delete `ServiceStatus` from the API completely (the definition and the endpoints), and also from model.go -- it's not used anywhere in the project, and your changes here make that struct irrelevant.

api/v1/gen/v1.yaml
meganokeefe: why is this enum needed? also i think service status can be more than just these https://github.com/sigma-dev/sigma/blob/master/model/model.go#L136

meganokeefe:
LGTM!

PR #172, opened 2017-06-21
Update docs to reflect new default manager intervals

kochman:
### Change log

Updated configuration page to reflect new defaults in PR #168.


helik:
lgtm^2

PR #173, opened 2017-06-21
Queue depth and back pressure detection

kochman:
### Change log

- Queue interface now has a Depth method that returns the number of messages in the queue
- Channel queue returns the length of the channel
- RabbitMQ queue asks for the number of messages in the queue via AMQP
- RabbitMQ queue sets QoS to only deliver one message at a time and leave the rest in the queue (to make them countable)
- Queue ack/nack functions can return errors
- Messages that are already being worked on are rejected

### Notes

Resolves the remaining half of #116.

I'm not sure at exactly what depth the service provider should stop sending services to the queue. Right now it's 100 but that's not based on anything. Maybe it should only enqueue if the queue depth is declining? Or maybe enqueue fewer services at a time while the queue depth is large?

~Waiting on #191.~

### TODO

- [x] Clean up back pressure detection in service_provider.go; kinda rough now
- [x] Get RabbitMQ queue depth through AMQP
- [x] Rip out rabbit-hole (RabbitMQ management API package)
- [x] Figure out why workers are stalling


manager/manager.go
```diff
@@ -176,14 +176,20 @@ func (m *manager) Start() error {
 						defer m.workerWg.Done()
 
 						m.updateService(svc)
-						m.svcConsume.Ack(svc)
+						err := m.svcConsume.Ack(svc)
```
adamschaub: Hold up. We might want to Ack before `updateService`. The latter can update the svc in the db. So it might be important to ack the message beforehand?

adamschaub: Discussed, ignored

adamschaub:
Works great, no holes as far as I can see. EXCEPT ONE. We may want to adapt the maximum queue depth based on the number of services to be queued. If we have 1000 services to queue, we will only ever queue up to 100 at a time. If the services returned from the database are always in the same order, we will only queue up to the first 100 services.

adamschaub:
Cool cool

Issue #122, opened 2017-06-02
API Object validation


adamschaub:
We need to centralize our validation of API objects. ATM, various API endpoints perform their own verification on incoming objects. 

For example, `AddService` allows the following object to be added:
```
{
  "name": "helloSvc",
  "plugin": "hello",
  "plugin_version": "v1",
  "input": [
    {
      "name": "Name",
      "type": "string",
      "value": "Unicorn"
    }
  ],
  "enabled": false
}
```

But `UpdateService` rejects the same object:
```
Site is required in a Service
```

Each endpoint should use the same validation before accepting an object.

PR #176, opened 2017-06-22
Api object validation

helik:
### Change log

- Centralize decoding request and validation of api object for Services, Plugins, Sites, and Bundles
- Add some tests for the validate methods

### Notes

The pattern for each of these is still very similar so it might be good at some point to make a generic method for checking similar things (ex: checking for invalid names).

Resolves: #122


dlb46:
Looks good to me!

meganokeefe:
can you `git pull origin master` to update your branch with the latest changes to `master`?

api/v1/server/server.go
```diff
@@ -1423,11 +1466,10 @@ func (s *Server) UpdateBundle(w http.ResponseWriter, r *http.Request) {
 		return
```
meganokeefe: not sure how it got here, but could you move the `UpdateBundle` function out of the bottom of the file and into the area where the other `Update` functions are?

api/v1/server/server.go
```diff
@@ -36,7 +36,10 @@ var (
 	ErrNoCert              = errors.New("Config needs a certificate.")
 	ErrNoKey               = errors.New("Config needs a key.")
 	ErrBadSiteID           = "Cannot change existing Site's ID"
+	ErrInvalidName         = "Invalid name"
+	ErrInvalidVersion      = "Invalid version"
 	ErrDuplicateName       = "Site already exists with the provided name"
```
meganokeefe: You didn't update this line, but could you change the error message from "Site" to "Object," since we use this error for validating other things like plugins, etc.?

api/v1/server/server_test.go
```diff
@@ -141,47 +142,38 @@ func TestValidateSite(t *testing.T) {
...
-			if validErr == nil {
-				t.Errorf("Test %d:\nsite=%v\nexpected err=%v, got err=%v\n", i+1, test.site, test.errExpect, validErr)
-			} else {
-				if validErr.Error() != test.errExpect.Error() {
-					t.Errorf("Test %d:\nsite=%v\nexpected err=%v, got err=%v\n", i+1, test.site, test.errExpect, validErr)
-				} else {
-					t.Logf("Test %d passed!", i+1)
-				}
-			}
+		if validErr != test.errExpect && ((validErr == nil && test.errExpect != nil) || (validErr != nil && test.errExpect == nil) || !strings.HasPrefix(validErr.Error(), test.errExpect.Error())) {
```
meganokeefe: logic behind switching from comparing errors to using `HasPrefix`?

meganokeefe:
Just a few comments, otherwise LGTM!

api/v1/server/server_test.go
helik: There are some places where the error is something like ErrSomething: %v, actualErrorMessage and that second part could vary, but the named error should be the same which is the prefix

helik:
updated!

PR #181, opened 2017-06-23
Secret Assets

mikeynap:
If an input/output is marked as "secret" it gets replaced with a one use URL.


api/v1/client/service_io.go
```diff
@@ -1,22 +1,24 @@
...
 	Type_ string `json:"type,omitempty"`
 
 	Value interface{} `json:"value,omitempty"`
 
 	From interface{} `json:"from,omitempty"`
 
 	Actual interface{} `json:"actual,omitempty"`
+
+	Secret    bool `json:"secret"`
+	SecretSet bool `json:"secret_set"` //TODO: REMOVE
```
meganokeefe: should this stay or go?

api/v1/server/server.go
```diff
@@ -21,6 +19,7 @@ import (
 	apiclient "github.com/sigma-dev/sigma/api/v1/client"
 	"github.com/sigma-dev/sigma/api/v1/convert"
 	"github.com/sigma-dev/sigma/auth"
+	"github.com/sigma-dev/sigma/auth/asset"
```
meganokeefe: there is also an empty `assets` folder in the root dir, can that folder be deleted?

auth/asset/asset.go
```diff
@@ -0,0 +1,97 @@
+package asset
+
+import "github.com/sigma-dev/sigma/model"
+
+const AssetType = "assetType"
+
+// Asset implements model.Asset
+// Each asset has "uses" that correspond to a token for that asset.
+// "Redeeming" subtracts one from that token and returns the underlying blob.
```
meganokeefe: I know very little about token/auth best practices -- is it idiomatic to se a token TTL based on # of uses?

auth/asset/asset.go
```diff
@@ -0,0 +1,97 @@
...
+
+type Asset struct {
+	model.MetaImpl
+
+	Id          string
+	AssetName   string
+	AssetBlob   []byte
```
meganokeefe: can you comment what `blob` is for, here

auth/asset/manager.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/sigma-dev/sigma/database"
+	"github.com/sigma-dev/sigma/model"
+)
+
+var (
+	ErrBadToken = errors.New("bad token")
+)
+
+type Manager interface {
+	Add(string, []byte, int) (model.Asset, string, error)
```
meganokeefe: comment the function headers for this interface?

auth/asset/manager.go
```diff
@@ -0,0 +1,129 @@
...
+
+type assetManager struct {
+	db database.SecretDB
+}
+
+func NewManager(db database.SecretDB) Manager {
+	return &assetManager{db}
+}
+
+func (m *assetManager) Add(name string, blob []byte, uses int) (model.Asset, string, error) {
```
meganokeefe: comment func header

cmd/sigma.go
```diff
@@ -167,3 +170,30 @@ func sigmaRun(cmd *cobra.Command, args []string) {
 		log.Info("Sigma Error Shutdown")
 	}
 }
+
+func InitTaskerAPIDefaults(gCfg *config.GlobalConfig) {
```
meganokeefe: this is nice

config.yaml
```diff
@@ -2,7 +2,8 @@
 Log:
   Level: info
 Manager:
-  OperationUpdateInterval: 5s
+  OperationUpdateInterval: 3s
+  UpdateInterval: 3ms
```
meganokeefe: did you want to keep this here?

config_nomad.yaml
```diff
@@ -0,0 +1,26 @@
+---
```
meganokeefe: remove this file, i think per @stephenrlouie we're just keeping `config.yaml` on master.

api/v1/server/server.go
```diff
@@ -1313,107 +1314,6 @@ func (s *Server) UpdateSite(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 }
-
-//validateSite is a helper that validates a Site before Add and Update
```
meganokeefe: make sure you've pulled master, validation stuff got merged

api/v1/server/server.go
```diff
@@ -1453,52 +1353,20 @@ func (s *Server) UpdateBundle(w http.ResponseWriter, r *http.Request) {
...
 	}
-	return false
-}
 
-// parse() is a helper for AddServiceOperationLogs/Progress/Outputs that converts from encoded base64 to string
-func parse(base string) (string, error) {
-	deHyphen := strings.Replace(base, "-", "", -1)
-	decoded, err := base64.StdEncoding.DecodeString(deHyphen)
-	parsed := string(decoded)
+	ass, err := s.assetManager.Redeem(tok)
```
meganokeefe: please rename, ha.

api/v1/server/validate.go
```diff
@@ -0,0 +1,161 @@
+package server
```
meganokeefe: ensure this is up-to-date with the validation kat did?

auth/asset/asset.go
```diff
@@ -0,0 +1,97 @@
...
+
+import "github.com/sigma-dev/sigma/model"
+
+const AssetType = "assetType"
+
+// Asset implements model.Asset
+// Each asset has "uses" that correspond to a token for that asset.
+// "Redeeming" subtracts one from that token and returns the underlying blob.
+
+type Asset struct {
```
meganokeefe: Reason why `Asset` is not part of `model.go`, `database/`?

auth/asset/asset.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (a Asset) Uses() int {
+	return a.AssetUses
+}
+
+func (a Asset) Blob() []byte {
+	return a.AssetBlob
+}
+
+func (a *Asset) AddUses(n int, s ...string) string {
```
meganokeefe: comment func header, what are these params

auth/asset/asset.go
```diff
@@ -0,0 +1,97 @@
...
+	return t
+}
+
+func (a *Asset) Redeem(s string) []byte {
+	if a.AssetUses == -1 {
+		return a.AssetBlob
+	} else if a.AssetUses == 0 {
+		return nil
+	}
+	for k := range a.AssetTokens {
```
meganokeefe: why have multiple tokens for an asset instead of just one?

auth/asset/manager.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	return ass, tok, nil
+}
+
+// GenFromToken takes a token, and adds a new single use token
+// to the underlying asset.
+func (m *assetManager) GenFromToken(tok string) (string, error) {
+	a := &Asset{}
+	var n string
+	sn, err := m.db.QuerySecrets(a, map[string]string{"Metadata." + tok: tok})
```
meganokeefe: remove period after `Metadata`? or if not accidental, why's it there?

auth/asset/manager.go
```diff
@@ -0,0 +1,129 @@
...
+	if len(sn) > 0 {
+		n = sn[0]
+	} else if err != nil {
+		return "", err
+	} else {
+		return "", ErrBadToken
```
meganokeefe: is `ErrBadToken` the right thing to throw here? it reaches this case if `len(sn) == 0` but no Query error was thrown -- there is no token to be `bad` then? 

shouldn't it be like, `ErrNoToken`?

auth/asset/manager.go
```diff
@@ -0,0 +1,129 @@
...
+
+	nTok := a.AddUses(1)
+	return nTok, m.db.UpdateSecret(a)
+}
+
+// Gen Token creates a token for an asset named `name`
+func (m *assetManager) GenToken(name string, uses ...int) (string, error) {
+	uses2Add := 1
+	if len(uses) != 0 {
+		uses2Add = uses[0]
```
meganokeefe: why are we letting `uses` be a slice if we only ever use the 1st element here?

tests/integration/manager/manager_test.go
```diff
@@ -21,10 +23,23 @@ import (
 var cli *apiclient.DefaultApi
 var flog *logrus.Logger
 var testDir string = "./_manager_integration_test_dir"
+var fooSecretKey string = "===SOMESSHKEY==="
+var fooSecretObj = `{"bah":"secret","blah":9}`
 
 const BASEURL = "http://127.0.0.1:24601/v1"
 
-func TestManager(t *testing.T) {
+func testManagerNomad(t *testing.T) {
```
meganokeefe: nit, capitalize `test`

task/docker.go
```diff
@@ -163,6 +190,40 @@ func (d *DockerTasker) Run(so *model.ServiceOperation) error {
 	return nil
 }
 
+func tarURLS(u map[string]string) ([]byte, error) {
```
meganokeefe: comment this function header

plugin/client/plugin_test.go
```diff
@@ -24,10 +24,16 @@ func TestMain(m *testing.M) {
 		return
 	}
 	testrun := m.Run()
-	os.RemoveAll(SaveDir)
+	//os.RemoveAll(SaveDir)
```
meganokeefe: uncomment?

plugin/client/plugin.go
```diff
@@ -61,17 +70,38 @@ func UnmarshalInputs(i interface{}) error {
 	return nil
 }
 
+func fillSecrets(s map[string]interface{}) {
```
meganokeefe: comment this

plugin/client/plugin.go
```diff
@@ -61,17 +70,38 @@ func UnmarshalInputs(i interface{}) error {
...
+		b, err := ioutil.ReadFile(p)
+		if err != nil {
+			return err
+		}
+		var m map[string]interface{}
+		if err = json.Unmarshal(b, &m); err == nil {
+			s[f.Name()] = m
+			return nil
+		}
+		s[f.Name()] = fmt.Sprintf("%v", string(b))
```
meganokeefe: comment this logic, so if there's an `unmarshal()` error you just return to return the string blob?

manager/manager.go
```diff
@@ -973,6 +1024,46 @@ func (m *manager) populateBundleInput(bs model.Service, in *model.ServiceDataIte
...
+}
+
+func tokenFromURL(a interface{}) (string, bool) {
+	s, ok := a.(string)
+	if !ok {
+		return "", false
+	}
+	parts := strings.Split(s, "/")
+	if len(parts) < 3 || parts[len(parts)-2] != "redeem" {
+		return "", false
```
meganokeefe: feel like `tokenFromUrl` should return an error instead of using the bool.

manager/manager.go
```diff
@@ -973,6 +1024,46 @@ func (m *manager) populateBundleInput(bs model.Service, in *model.ServiceDataIte
 	return nil
 }
 
+func (m *manager) processSecret(svcName, itemName string, v *model.ServiceDataItem, val interface{}) error {
```
meganokeefe: comment this func header

tests/integration/manager/manager_test.go
mikeynap: It's purposefully not capitalized -- we don't want this to run automatically -- once Nomad works correctly we'll have it run.

manager/manager.go
```diff
@@ -551,23 +557,45 @@ func collectOutput(svc *model.Service, plugin model.Plugin, output map[string]in
...
 	for i, item := range svc.Output {
+		v, ok := out[i]
 		output := svc.Output[i]
+		wasSet := output.Set
 		output.Set = false
-		if v, ok := out[i]; ok {
-			if conv, ok := model.ConvertIO(v, output.Type, plugin); ok {
+		if ok { // if task output exists.
+			if item.Secret {
+				if item.SecretSet {
```
meganokeefe: why is `SecretSet` needed, can't you just check if `Secret` is non-null?

auth/asset/asset.go
```diff
@@ -0,0 +1,105 @@
...
+// Each asset has "uses" that correspond to a token for that asset.
+// "Redeeming" subtracts one from that token and returns the underlying blob.
+
+type Asset struct {
+	model.MetaImpl
+
+	Id          string
+	AssetName   string
+	AssetBlob   []byte
+	AssetTokens map[string]int
```
meganokeefe: why are tokens `ints` vs. strings?

auth/asset/manager.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+var (
+	ErrBadToken = errors.New("bad token")
+)
+
+type Manager interface {
+	Add(string, []byte, int) (model.Asset, string, error)
+	GenToken(string, ...int) (string, error)
+	GenFromToken(string) (string, error)
```
meganokeefe: motion to rename this function, I don't think it's descriptive of what the function actually does / is confusingly similar to `GenToken`. maybe change to `AddAssetToken`? or `CreateAndAddToken`?

model/model.go
```diff
@@ -286,3 +292,21 @@ type Bundle struct {
 	Input    map[string]PluginDataItem
 	Output   map[string]PluginDataItem
 }
+
+// Secrets are guarenteed to have type info set.
+type Secret interface {
```
meganokeefe: Why are you defining interfaces here instead of structs?

auth/asset/manager.go
```diff
@@ -0,0 +1,129 @@
...
+		return nil, database.ErrNotFound
+	}
+	ass, ok := sec.(*Asset)
+	if !ok {
+		return nil, database.ErrNotFound
+	}
+	return ass, nil
+}
+
+// Gets asset with token, decrements it's use counter and returns the blob
```
meganokeefe: it's -> its

auth/asset/manager.go
meganokeefe: check `err` here

task/docker.go
```diff
@@ -146,8 +149,28 @@ func (d *DockerTasker) Run(so *model.ServiceOperation) error {
 	if err := os.MkdirAll(dir, 0775); err != nil {
 		return err
 	}
+	if err := os.MkdirAll(path.Join(dir, "secrets"), 0775); err != nil {
+		return err
+	}
 
-	jInput, err := json.Marshal(model.ServiceDataItemsToMap(so.Service.Input))
+	inputMap, secretMap := model.ServiceDataItemsToMaps(so.Service.Input)
```
meganokeefe: NIT, can you move this inputs/secrets stuff into a helper, to prevent `Run()` from getting too bulky?

task/docker.go
```diff
@@ -146,8 +149,28 @@ func (d *DockerTasker) Run(so *model.ServiceOperation) error {
...
+	sDir := path.Join(dir, "secrets")
+	for k, v := range secretMap {
+		res, err := http.Get(v)
+		if err != nil {
+			return err // todo... actual err?
+		} else if res.StatusCode == http.StatusUnauthorized {
+			return errors.New("unable to get secret")
```
meganokeefe: should you add an "unauthorized" message to the err here?

task/docker.go
```diff
@@ -146,8 +149,28 @@ func (d *DockerTasker) Run(so *model.ServiceOperation) error {
...
 	}
+	if err := os.MkdirAll(path.Join(dir, "secrets"), 0775); err != nil {
+		return err
+	}
 
-	jInput, err := json.Marshal(model.ServiceDataItemsToMap(so.Service.Input))
+	inputMap, secretMap := model.ServiceDataItemsToMaps(so.Service.Input)
+	sDir := path.Join(dir, "secrets")
+	for k, v := range secretMap {
+		res, err := http.Get(v)
```
meganokeefe: can you comment what's going on here

database/couchbase.go
```diff
@@ -843,15 +846,15 @@ func (cb *Couchbase) DeleteSecret(s Secret) error {
...
+func (cb *Couchbase) UpdateSecret(s model.Secret) error {
 	if s == nil {
 		return ErrEmptyID
 	}
 	s.Set("Name", s.Name())
 	return cb.update(s.ID(), s.Type(), s)
 }
 
-func (cb *Couchbase) QuerySecrets(ss Secret, filters map[string]string) ([]string, error) {
+func (cb *Couchbase) QuerySecrets(ss model.Secret, filters map[string]string) ([]string, error) {
```
meganokeefe: I feel that the params  for `QuerySecrets` should align with the other db.Query functions (single struct param). Especially with the artificial "one param should be empty" thing. Could separate into two functions if you need to query by `filters`?

auth/asset/manager.go
```diff
@@ -0,0 +1,129 @@
...
+func NewManager(db database.SecretDB) Manager {
+	return &assetManager{db}
+}
+
+func (m *assetManager) Add(name string, blob []byte, uses int) (model.Asset, string, error) {
+	unlimited := false
+	if uses == -1 {
+		unlimited = true
+	}
+	ass := NewAsset(name, blob, unlimited)
```
meganokeefe: please rename

database/nulldb.go
```diff
@@ -26,7 +29,7 @@ type NullDB struct {
 	SitesName                map[string]*model.Site
 	BundlesName              map[string]*model.Bundle
```
meganokeefe: nit, remove whitespace

database/nulldb.go
```diff
@@ -44,14 +47,14 @@ func NewNullDB(casFails bool) *NullDB {
 		make(map[string]model.ServiceOperation),
 		make(map[string]model.ServiceOperationLogs),
 		make(map[string]model.Bundle),
-		make(map[string]Secret),
+		make(map[string][]byte),
```
meganokeefe: why `[]byte` and not `Secret`?

database/nulldb.go
```diff
@@ -87,9 +92,12 @@ func (n *NullDB) GetService(id string) (model.Service, error) {
 			r = *r2
 		}
 	}
+
 	if ok {
-		r.Set("cas", 949328)
-		return r, nil
+		var rc model.Service
+		Copy(&rc, r)
```
meganokeefe: what is going on here with the `Copy`

database/nulldb.go
```diff
@@ -682,29 +699,31 @@ func (n *NullDB) CreateLock(ID string, body string, ttl time.Duration) error {
...
 
 // SECRETS
-func (n *NullDB) InsertSecret(o Secret) (string, error) {
+func (n *NullDB) InsertSecret(o model.Secret) (string, error) {
 	n.Lock()
 	defer n.Unlock()
 	o.ID(model.GenerateUUID())
-
-	n.Secrets[o.ID()] = o
+	d, _ := json.Marshal(o)
```
meganokeefe: why

database/nulldb.go
```diff
@@ -717,29 +736,61 @@ func (n *NullDB) DeleteSecret(s Secret) error {
...
-		n.SecretsName[s.Name()+"_"+s.Type()] = s
+		n.SecretsName[s.Name()+"_"+s.Type()] = d
 		n.NWithName[s.Name()]++
 	}
 	return nil
 }
+
+// Make a deep copy from src into dst.
+// https://github.com/getlantern/deepcopy
+func Copy(dst interface{}, src interface{}) error {
```
meganokeefe: is this for race condition fixing purposes? why is this needed

manager/manager.go
```diff
@@ -254,13 +257,14 @@ func (m *manager) processBundleService(s model.Service) {
 func (m *manager) updateService(serviceID string) {
 	//Check if operation exists
 	op, err := m.db.GetServiceOperation(operationID(serviceID))
-	if err != nil && err != database.ErrNotFound {
+	if err != nil && (err != database.ErrNotFound && err != database.ErrEmptyID) {
```
meganokeefe: why ignore `ErrEmptyID` here? isn't that a valid error?

tests/integration/manager/manager_test.go
```diff
@@ -117,11 +144,14 @@ func testServices(tt *testing.T) error {
...
 			},
 			map[string]interface{}{
-				"url":        "127.0.1.2",
-				"foo-output": []string{"Hello, ", "World"},
-				"other":      "manualvalue",
-				"fooObj":     map[string]interface{}{"blah": 3, "bah": "caz"},
+				"url":           "127.0.1.2",
+				"foo-output":    []string{"Hello, ", "World"},
+				"other":         "manualvalue",
+				"secretKey":     "===SOMESSHKEY===",
```
meganokeefe: use the `fooSecretKey` var here?

auth/asset/asset.go
mikeynap: It's the []byte version of the asset being stored.

auth/asset/manager.go
mikeynap: yeah good call

cmd/sigma.go
mikeynap: thx :)

auth/asset/asset.go
mikeynap: Because model.go/database use an interface called "Secret" for storing secrets.
We have two types of "Secrets" so far -- Users and Assets. 

Instead of having to write function definitions for both, they both use the same "Secrets" interface defined in model.

auth/asset/asset.go
mikeynap: What, n and s... aren't clean enough :p lolol. Will do.

auth/asset/asset.go
mikeynap: I debated this... this way is "more secure". 
The way your describing looks like this:
Say we have a server service with a "SSHKey" secret output. A lot of services would like to use that. Say 10 services are interested. This server's "SecretKey" field is t=http://...v1/.../token123

So now each of the 10 services has the same token `t`. This token (before it is used by the Tasker) has 10 uses. If anyone of them gets this token somehow intercepted, the attacker has 10 times as long as normal to try and use the token. 

In the way I have it written, the parent service uses it's token to generate single use tokens for anything that wants it.

auth/asset/manager.go
mikeynap: It wants to Filter on Metadata

auth/asset/manager.go
mikeynap: One would argue that having no token is a bad token... but I see your point :p

auth/asset/manager.go
mikeynap: It's an optional parameter. Which defaults to 1.

plugin/client/plugin.go
mikeynap: yeah I added a comment -- if the input is a json object, try to unmarshal it. 
If it is not, then use the contents of the file as-is.

mikeynap: Useful for SSH keys

auth/asset/asset.go
mikeynap: it's a map of token->uses.

auth/asset/manager.go
mikeynap: good call.

model/model.go
mikeynap: Multiple parts of the project will be using secrets -- the DB interface uses interfaces to allow any type of secret. 

Now we have "Assets" and "Users", both with conform to "Secret".

auth/asset/manager.go
mikeynap: it is what it is.
jk.

auth/asset/manager.go
mikeynap: we do?

task/docker.go
mikeynap: yepppp

task/docker.go
mikeynap: sure.

database/couchbase.go
mikeynap: It will align once my other PR hits lol. It adds filters to to all objects.

auth/asset/manager.go
mikeynap: booo

database/nulldb.go
mikeynap: Due to stupid ass race conditions. Makes it much easier (and without races).

database/nulldb.go
mikeynap: This is the fix for race conditions. Copy does a deep copy of the objects before they are returned.

mikeynap: As a side effect, it also makes it behave more like Couchbase.

database/nulldb.go
mikeynap: Explained above.

database/nulldb.go
mikeynap: yes. see above.

manager/manager.go
mikeynap: If the service operation is not found, then that means we have to create one below in the function.

This snippet took me like an hour to understand yesterday lol. I THINK this is right.

tests/integration/manager/manager_test.go
mikeynap: sher.

api/v1/server/validate.go
mikeynap: it is, I copied and pasted it.

auth/asset/asset.go
adamschaub: The real question. Why not just check if the token exists? Why the loop

auth/asset/manager.go
```diff
@@ -0,0 +1,138 @@
...
+	// Generate a new token for an asset named `name`
+	GenToken(name string, optionalNUses ...int) (string, error)
+
+	// Generate a new token from an old token (without consuming the old token).
+	GenFromToken(tok string) (string, error)
+
+	// Redeem a token for an asset. Consumes a use of the token.
+	Redeem(tok string) (model.Asset, error)
+
+	// Delete an asset named `name` from the secretDB
```
adamschaub: Name or the token?

auth/asset/manager.go
```diff
@@ -0,0 +1,138 @@
...
+type assetManager struct {
+	db database.SecretDB
+}
+
+func NewManager(db database.SecretDB) Manager {
+	return &assetManager{db}
+}
+
+func (m *assetManager) AddAsset(name string, blob []byte, uses int) (model.Asset, string, error) {
+	unlimited := false
```
adamschaub: `unlimited := uses == -1`

auth/asset/manager.go
```diff
@@ -0,0 +1,138 @@
...
+	ass, ok := sec.(*Asset)
+	if !ok {
+		return nil, database.ErrNotFound
+	}
+	return ass, nil
+}
+
+// Gets asset with token, decrements its use counter and returns the blob
+func (m *assetManager) Redeem(tok string) (model.Asset, error) {
+	a := &Asset{}
```
adamschaub: Small nit. Given you do this same  thing in `GenFromToken`, maybe make it's own func? If it is only these two locations, probably ignore...

cmd/sigma.go
adamschaub: Hm, maybe worth considering defaults in the config pkg...

auth/asset/asset.go
adamschaub: Can you doc that for this struct? Just reference what interface it fufills.

database/nulldb.go
adamschaub: Both databases behave in the same way?!? *gasp*

database/nulldb.go
```diff
@@ -400,13 +416,13 @@ func (n *NullDB) QueryServices(o model.Service) ([]model.Service, error) {
...
 	mdMatch := true
-	for _, v := range n.Services {
+	for i, v := range n.Services {
 		for k, mdv := range o.Metadata {
 			mdMatch = mdMatch && mdv == v.Metadata[k]
 		}
 
 		if mdMatch && (o.Id == "" || o.Id == v.Id) && (o.SaveDir == "" || o.SaveDir == v.SaveDir) && (o.Name == "" || o.Name == v.Name) && (o.PluginName == "" || o.PluginName == v.PluginName) && (o.PluginVersion == "" || o.PluginVersion == v.PluginVersion) {
-			r = append(r, v)
+			r = append(r, n.Services[i])
```
adamschaub: For my curiosity, why did these have to change?

database/nulldb.go
adamschaub: Is the same true for the rest of the objects? Couldn't you use `Copy` to avoid the same race? Or is the deeply nested pointer issue you ran into earlier.

database/nulldb.go
```diff
@@ -717,29 +735,61 @@ func (n *NullDB) DeleteSecret(s Secret) error {
...
-func (n *NullDB) QuerySecrets(s Secret, filters map[string]string) ([]string, error) {
+func (n *NullDB) QuerySecrets(s model.Secret, filters map[string]string) ([]string, error) {
+	n.Lock()
+	defer n.Unlock()
 	var secs []string
+secretLoop:
 	for _, sec := range n.Secrets {
-		if sec.Name() == s.Name() || sec.Type() == s.Type() {
-			secs = append(secs, sec.Name())
+		ts := reflect.New(reflect.TypeOf(s).Elem()).Interface().(model.Secret)
```
adamschaub: What is this accomplishing?

manager/manager.go
adamschaub: If the operation doesn't exist, there would be no ID right? Why need to check for `ErrEmptyID` as well?

manager/manager.go
```diff
@@ -442,14 +448,16 @@ func (m *manager) finishOperation(op model.ServiceOperation) (retErr error) {
 	case model.ServiceOperationDeploy:
 		svc.LastStatus = time.Now()
 		if len(op.Output) == 0 {
-			log.Warnf("%v, operation=%s", task.ErrServiceOutputNotAvailable, op.Id)
+			if op.Operation == model.ServiceOperationDeploy {
```
adamschaub: On this note, maybe it's just OKAY to nave no outputs. Surely not every service will.

manager/manager.go
```diff
@@ -591,17 +600,38 @@ func collectOutput(svc *model.Service, plugin model.Plugin, output map[string]in
 
 	for i, item := range svc.Output {
 		output := svc.Output[i]
+		wasSet := output.Set
```
adamschaub: Every time I look at this function, I want to barf

adamschaub: And it's not getting better

manager/manager.go
adamschaub: I'd disagree with this. Either you can get the token from the string or not. No need to define an error.

manager/manager.go
```diff
@@ -152,21 +153,21 @@ func (m *manager) Start() error {
 		}
 	}
 
-	ticker := time.NewTicker(m.updateInterval)
+	ticker := time.After(m.updateInterval)
```
adamschaub: I'm guessing this was an issue because ticker wasn't taking processing time into account?

model/convert.go
```diff
@@ -88,14 +88,14 @@ func convertPluginIO(a interface{}, t string, plug Plugin, entry bool) (interfac
 	}
 	// If type is object, convert subfields
 	var outObj map[string]interface{}
-	if outObj, ok = out.(map[string]interface{}); !ok {
+	if outObj, ok = out.(map[string]interface{}); !ok || outObj == nil {
```
adamschaub: why the need for a nil check? An object should technically be allowed to be nil

model/model.go
```diff
@@ -183,16 +183,20 @@ func (s Service) IsPartOfBundle() bool {
...
 
 // lists must be of type 'list/'
-func ServiceDataItemsToMap(in map[string]ServiceDataItem) map[string]interface{} {
+func ServiceDataItemsToMaps(in map[string]ServiceDataItem) (map[string]interface{}, map[string]string) {
 	sdimap := make(map[string]interface{})
+	secretMap := make(map[string]string)
 	for name, i := range in {
-		if i.Set {
+		asString, ok := i.Actual.(string)
+		if i.Secret && ok {
```
adamschaub: `if asString, ok := i.Actual.(string); i.Secret && ok`

plugin/client/plugin.go
```diff
@@ -70,6 +78,29 @@ func unmarshalInputFile(i interface{}, src string) error {
...
+	filepath.Walk(secretSavedir, func(p string, f os.FileInfo, err error) error {
+		if f == nil || f.IsDir() {
+			return nil
+		}
+		b, err := ioutil.ReadFile(p)
+		if err != nil {
+			return err
+		}
+		// If the file contains a JSON object...
+		var m map[string]interface{}
```
adamschaub: All secrets will be a JSON object? I had imagined that they would be individual files. Say, one for each cert/key/token.

plugin/client/plugin_test.go
```diff
@@ -28,6 +28,12 @@ func TestMain(m *testing.M) {
 	os.Exit(testrun)
 }
 
+func TestSecret(t *testing.T) {
+	s := make(map[string]interface{})
+	fillSecrets(s)
+	t.Log(s)
+}
```
adamschaub: WHAT A THOROUGH TEST

task/docker.go
```diff
@@ -162,16 +165,21 @@ func (d *DockerTasker) Run(so *model.ServiceOperation) error {
...
 		return err
 	}
 
 	if err = ioutil.WriteFile(path.Join(dir, "input"), jInput, 0666); err != nil {
 		return err
 	}
-
-	ojInput, err := json.Marshal(model.ServiceDataItemsToMap(so.PreviousInput))
+	oldInMap, _ := model.ServiceDataItemsToMaps(so.PreviousInput)
+	ojInput, err := json.Marshal(oldInMap)
```
adamschaub: Secrets are also considered inputs, so we'll have to run updates when secrets are updated too? We have to think about this area more in depth.

task/docker.go
```diff
@@ -432,6 +441,28 @@ func readAll(reader io.Reader) ([]byte, error) {
 	return b, nil
 }
 
+func processSecrets(sMap map[string]string, saveDir string) error {
+	if err := os.MkdirAll(saveDir, 0775); err != nil {
```
adamschaub: Reckon we can stick with 6's

manager/manager.go
mikeynap: yepppp

database/nulldb.go
mikeynap: yeah it's the deeply nested garbage.

manager/manager.go
mikeynap: I was getting random ErrEmptyIDs back... maybe we should look into why that was happening instead of me just bandaging it here.

auth/asset/manager.go
mikeynap: By Token.

auth/asset/manager.go
mikeynap: 🤦‍♂️

auth/asset/manager.go
mikeynap: was a good idea any, so I did it.

task/docker.go
mikeynap: not sure that works with directories.

mikeynap:
99!

mikeynap:
100!

adamschaub:
@mikeynap Fixed some of the lingering issues we saw w.r.t. inputs as secrets, running ops multiple times, etc. I'm a little perplexed with some of the merge issues though, could use help updating with master.

PR #182, opened 2017-06-23
DockerTasker filesystem monitoring

### LINKED ISSUE #174: DockerTasker should be standardized with Nomad (Ship)
### problem 

The way we do Progress/Outputs Shipping between the two taskers is different right now -- 

NomadTasker has a shipper script that it runs as an exec sidecar task for every job. (Ie. a curl script for progress/output is run alongside every plugin container). Nomad controls when this `curl` happens using its own scheduler.

But DockerTasker does these `curl`s on-demand-- specifically, manager calls `DockerTasker.Ship()` at the configurable service update interval, which gets progress/output into the DB, and then on the next-loop around, manager has those fields at its disposal. 

### solution 

Because we're locked into this pattern for Nomad, we should change DockerTasker to, instead of having `tasker.Ship`, to spawn a goroutine that monitors the filesystem's progress/output for changes, and `curl`s to the sigma API anytime there's a change. 

This goroutine should also (similar to nomad) gracefully shutdown when `ContainerStop` happens, `curl`ing the progress/outputs one last time. 

Basically remove the `Ship()` function from the tasker interface completely.

meganokeefe:
Added support in DockerTasker for on-the-fly POSTs of serviceoperation `Progress` and `Output`, which standardizes the DockerTasker shipping behavior with that of [NomadTasker](https://github.com/sigma-dev/sigma/blob/master/task/nomad.go#L192). 

(AKA, manager no longer has to tell DockerTasker when to POST.)

Closes #174.  

### Changelog 

- [x] Vendored [watcher](https://github.com/radovskyb/watcher), a filesystem monitoring package
- [x] Removed manager's call to `DockerTasker.Ship()`
- [x] Removed the `Ship` function from the tasker interface, updated the mock tasker 
- [x] Now when we call `NewDockerTasker`, we spawn a `Watcher.Start()` that monitors the savedir (and its current and future subdirectories) for file `create`+`write` operations. The watcher then triggers an API post for progress/outputs using the files in the relevant subdirectory. 
- [x] integration test to verify that `ship()` can post to the API 
- [x] added `tasker.Shutdown` to the tasker interface, which manager calls on manager shutdown. This is where the Watcher is stopped.

### Notes 

Initially i was spawning a separate `Watcher` for every `DockerTasker.Run` and storing the associated serviceGroups in a map- but now I have a single `Watcher`.

### TODO 

Not totally sure how to approach the shutdown of the Watcher. Because we're starting it up in `NewDockerTasker`, we don't have a `Tasker.Stop` function to match that / an apparent place to put the `svc.Kill()`. So right now I'm just letting it die. Advice about where to put the `Kill` is appreciated.


manager/manager.go
```diff
@@ -399,11 +399,8 @@ func (m *manager) startOperation(op model.ServiceOperation) {
 }
 
 func (m *manager) statusOperation(op model.ServiceOperation) {
-	stat := ""
-	err := m.tasker.Ship(&op)
-	if err != nil {
-		log.WithError(err).Warnf("Ship failed for operation %s", op.Id)
-	}
+	var stat string
+	var err error
```
helik: do you need this line? It doesn't look like you use err outside this block

helik:
"I have a defer watcher.Close() inside the watchAndShip function that lets the watcher gracefully shutdown, and a debug statement after that, but that statement never prints, making me think that the watcher is never gracefully shut down." --> where is this debug statement?

helik:
To me, it looks like the go routines you start (both the function watchAndShip and the anonymous one inside of watchAndShip) don't ever shut down.

The anonymous one will never shut down because it is in an infinite for select loop with no case that breaks it out. The watchAndShip one will also never shut down because the done channel is never signaled so it is stuck on the <-done line. 

From my brief look at it, it seems that you don't need to spawn multiple go routines. You could just call watchAndShip and spawn the anonymous one which is what is watching in a loop anyways. Is there a reason that watchAndShip is waiting on the done channel? If not, it can just exit.

If the desired behavior is to stop the go routine when a shut down happens (is this a Sigma shutdown?), then you will need a way for the go routine to know when that happens. With that, you can add a case on the select and break out of the for select which will then exit the go routine.

task/docker.go
```diff
@@ -234,33 +235,40 @@ func (d *DockerTasker) Status(so *model.ServiceOperation) (string, error) {
 // watchAndShip monitors the serviceOperation savedir and POSTs progress/outputs to sigma API
 // every time the savedir has a "write" operation (file create / file update)
 func (d *DockerTasker) watchAndShip(so *model.ServiceOperation, dir string) {
-	log.Debugf("Watching and shipping for %s", so.Id)
+	log.Info("---- entered watchAndship ---------")
+	//setup
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+	quit := make(chan os.Signal, 1)
```
helik: does this mean the os will send on the quit channel when the program wants to shut down? or if not, who sends on this channel?

helik:
It looks as though you got rid of both "go" statements. If you want to do the for select directly in watchAndShip, then the call to watchAndShip should probably spawn a go routine. If this is the case, then you don't need the wait group since it will already be waiting for the for select to continue.

Otherwise, you should add a go func around the for select like you had before. If you do this, then you should not have a wait group since it will block the method ending. Instead of waiting for the for select to finish, you could just add the defer watcher.Close() to before the for select (inside the anonymous go routine) and add a close(quit) after the quit channel has been signaled. 

Also, I'm not sure how often the watcher.Events channel will signal, but it might be good to put the quit case first so it will exit immediately.

task/watcher.go
```diff
@@ -0,0 +1,54 @@
...
+		select {
+		case event := <-w.Watch.Events:
+			if event.Op&fsnotify.Write == fsnotify.Write || event.Op&fsnotify.Create == fsnotify.Create {
+				shipErr := w.DT.ship(w.ServiceOp)
+				if shipErr != nil {
+					log.WithError(shipErr).Warnf("Ship operation failed for serviceOperation=%s", w.ServiceOp.Service.Name)
+				}
+			}
+		case <-w.Quit:
+			break ctrl_loop
```
adamschaub: could just log and return in this case, and remove the label

adamschaub: Also consider @helik's point about reordering. If fsnotify is conservative in events being fired, I wouldn't worry about it.

helik:
lgtm! lgtm

api/v1/server/server.go
```diff
@@ -534,18 +534,21 @@ func (s *Server) AddServiceOperationProgress(w http.ResponseWriter, r *http.Requ
 	vars := mux.Vars(r)
 	serviceName, ok := vars["service-name"]
 	if !ok {
+		log.Error("NO NAME")
```

api/v1/server/server.go
```diff
@@ -534,18 +534,21 @@ func (s *Server) AddServiceOperationProgress(w http.ResponseWriter, r *http.Requ
...
 		http.Error(w, "no service-name in AddServiceOperationProgress request", http.StatusBadRequest)
 		return
 	}
 	serviceOpId, ok := vars["service-operation-id"]
 	if !ok {
+		log.Error("NO ID")
```

api/v1/server/server.go
```diff
@@ -534,18 +534,21 @@ func (s *Server) AddServiceOperationProgress(w http.ResponseWriter, r *http.Requ
...
 		http.Error(w, "no service-operation-id in AddServiceOperationProgress request", http.StatusBadRequest)
 		return
 	}
 
 	// decode and parse progress into string
 	var apiObj map[string]interface{}
 	if err := decoder.Decode(&apiObj); err != nil {
+		log.Error(err)
```

api/v1/server/server.go
```diff
@@ -567,17 +571,20 @@ func (s *Server) AddServiceOperationProgress(w http.ResponseWriter, r *http.Requ
 		var op2Update model.ServiceOperation
 		var err error
 		if op2Update, err = s.db.GetServiceOperation(serviceOpId); err != nil {
+			log.Error(err)
```
adamschaub: remove

api/v1/server/server.go
```diff
@@ -567,17 +571,20 @@ func (s *Server) AddServiceOperationProgress(w http.ResponseWriter, r *http.Requ
...
 			http.Error(w, err.Error(), http.StatusNotFound)
 			return http.StatusNotFound, err
 		}
 		op2Update.Progress = parsedProg
 		if err = s.db.UpdateServiceOperation(op2Update); err != nil {
+			log.Error(err)
```
adamschaub: remove

api/v1/server/server.go
```diff
@@ -567,17 +571,20 @@ func (s *Server) AddServiceOperationProgress(w http.ResponseWriter, r *http.Requ
...
 			return http.StatusBadRequest, err
 		}
 		return http.StatusOK, nil
 	}
 
 	if rc, err := tryUpdate(updateFunc, UpdateRetryTimeout); err != nil {
+		log.Error(err)
```
adamschaub: remove

task/docker.go
```diff
@@ -91,9 +93,10 @@ func NewDockerTasker(cfg DockerConfig, cl *http.Client, fluentUrl string) (*Dock
 	} else {
 		t.saveDir = absPath
 	}
-
-	os.MkdirAll(t.saveDir, 0777)
-
+	err = os.MkdirAll(t.saveDir, 0777)
+	if err != nil {
+		log.Error(err)
```
adamschaub: Return error

adamschaub: Instead of logging

task/docker.go
```diff
@@ -184,19 +200,35 @@ func (d *DockerTasker) Run(so *model.ServiceOperation) error {
...
 
 //Remove will remove the docker container associated with the Service
 func (d *DockerTasker) Remove(so *model.ServiceOperation) error {
+	// Check operation
 	if so == nil || so.Service.Id == "" {
 		return ErrNilServiceOperation
 	}
+
+	// Ship one more time before destroy
+	time.Sleep(time.Second * 3)
```
adamschaub: Why the sleep?

task/docker.go
```diff
@@ -184,19 +200,35 @@ func (d *DockerTasker) Run(so *model.ServiceOperation) error {
...
+	err := d.ship(so)
+	if err != nil {
+		log.Warn(err)
```
adamschaub: Return instead

task/watcher.go
```diff
@@ -0,0 +1,53 @@
...
+	Dir       string
+	Quit      chan struct{}
+	DT        *DockerTasker
+}
+
+func (w *Watcher) Start() error {
+	var err error
+	w.Watch, err = fsnotify.NewWatcher()
+	if err != nil {
+		log.Error(err)
```
adamschaub: remove

task/watcher.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	err = w.Watch.Add(w.Dir)
+	if err != nil {
+		log.Error(err)
+		return err
+	}
+	for {
+		select {
+		case <-w.Quit:
+			log.Debugf("Shutting down FS monitor for %s", w.ServiceOp.Id)
```
adamschaub: Remove

api/v1/server/server.go
```diff
@@ -642,6 +639,7 @@ func (s *Server) AddServiceOperationOutputs(w http.ResponseWriter, r *http.Reque
 		return http.StatusOK, nil
 	}
 	if rc, err := tryUpdate(updateFunc, UpdateRetryTimeout); err != nil {
+		log.Error(err)
```
adamschaub: REMOVE

task/docker.go
```diff
@@ -322,7 +349,7 @@ func (d *DockerTasker) postShipper(so *model.ServiceOperation, payload []byte, e
 		}
 		defer resp.Body.Close()
 	} else {
-		// otherwise use DockerTasker's client (needed for testing / mock client)
+		// otherwise use DockerTasker's provided client (needed for testing / mock client)
 		resp, err := d.HTTPClient.Do(req)
 		if err != nil {
 			log.Error(err)
```
adamschaub: REMOVE

task/watcher.go
```diff
@@ -0,0 +1,50 @@
...
+
+func (w *Watcher) Start() error {
+	var err error
+	w.Watch, err = fsnotify.NewWatcher()
+	if err != nil {
+		return err
+	}
+	err = w.Watch.Add(w.Dir)
+	if err != nil {
+		log.Error(err)
```
adamschaub: REMOVE

PR #184, opened 2017-06-23
Bug #170 Delete Cycle Detect

### LINKED ISSUE #170: api/v1/server/server.go CycleDetect not thread safe
##### Description

This function is not thread safe as numerous functions can be running `CycleDetect` at once. Two functions that will be in the same tree could be checking for cycles but not detect them. The combination of those new assignments could cause a cycle.

**Example:**
`SiteA` - no parents
`SiteB` - Parent: `SiteP`

I make a request to assign `SiteA`'s parent to `SiteB` and at the same time I assign `SiteB`'s parent to be `SiteA`. `SiteA` runs CycleDetect, and sees B's parent is P. `SiteB` runs CycleDetect and sees A has no parents. So they both are validated and assign their new parents and we have a cycle. `SiteA` <-> `SiteB`

##### Steps to reproduce

Hard to reproduce since it comes down to timing of update calls. I guess try to replicate what I described, but I wouldn't recommend since you have to do it many many times for it to work out. 

##### Suggest a Fix

[DB Locks ](https://github.com/sigma-dev/sigma/blob/master/database/couchbase.go#L798)

- Only allow one function to enter CycleDetect at a time. (fastest but least concurrent way)
- Only allow updates if the parent or anyone in its parent tree is *not* being updated. So we will lock down single trees not everything. (Requires locks per site, might be more complicated to get right, but will allow for more concurrency if tree roots differ.)

dlb46:
Fix #170 
### Change log

Deleted CycleDetect function.  Added this cycle check into gatherChildSites function.

Reason for helper function it help us know the first time we are calling gatherChildSites, because it is a recursive function and we need to keep track of the root.

### Notes

This method works but if anyone has a better/clearer way to implement this let me know!


stephenrlouie:
Code looks good, but we should remove the DetectCycle code in server.

# TODO: 
- [ ] Clean up [CycleDetect](https://github.com/sigma-dev/sigma/blob/bd2fe0c7a86bb3ce372ceaf9fb151918b46042cb/api/v1/server/server.go#L1475-L1496) and any references to it.

manager/manager.go
```diff
@@ -839,7 +839,16 @@ func (m *manager) newOperation(svc *model.Service, plugin model.Plugin, opType s
 }
 
 // recursive helper for newOperation, goes down the site tree to get all child Sites plus myself
+//the reason for the helper is so we know if we are calling gatherChildSitesWorker for the first time (with the root)
 func (m *manager) gatherChildSites(parentName string) ([]model.Site, error) {
+	children, err := m.gatherChildSitesWorker("", parentName, true)
```
stephenrlouie: you can pass in the originName here

manager/manager.go
```diff
@@ -850,13 +859,19 @@ func (m *manager) gatherChildSites(parentName string) ([]model.Site, error) {
...
 	}
-	//base case - No children to recurse down
-	if len(children) == 0 {
+	//base case - No children to recurse down or completed the cycle back to myself
+	if len(children) == 0 || originName == parentName && !isFirst{
 		return allChildSites, nil
 	}
+	//necessary to make sure we keep the correct origin name
+	//if isFirst is true then we know it is the first go around, so set originName to parentName
+	if isFirst {
```
stephenrlouie: can remove if we pass in origin name

PR #185, opened 2017-06-23
Cookies!

mikeynap:
Added Cookies. The API token gets written to a cookie.


auth/tokenizer/jwt.go
```diff
@@ -15,16 +15,16 @@ var (
...
 type JWTTokenizer struct {
 	secret []byte
 	method jwt.SigningMethod
-	ttl    uint32
+	ttl    int64
 }
 
 // TODO: SecretFetcher()???
-func NewJWT(secret string, ttl ...uint32) *JWTTokenizer {
+func NewJWT(secret string, ttl ...int64) *JWTTokenizer {
```
kochman: Why variadic if only the first element is used?

api/v1/server/middleware.go
```diff
@@ -24,7 +24,11 @@ func (s *Server) Auth(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		auth := r.Header.Get("Authorization")
 		token := strings.TrimPrefix(auth, "Bearer ")
-		if tl := len(token); tl == 0 && len(auth) != 0 {
+		cookie, err := r.Cookie(loginCookieName)
+
+		if err == nil {
+			token = cookie.Value
+		} else if tl := len(token); tl == 0 && len(auth) != 0 {
```
mikeynap: lol yeah

auth/tokenizer/jwt.go
mikeynap: Optional value. NewJWT("secret") uses the default TTL.

Issue #180, opened 2017-06-23
ditto service always updating


stephenrlouie:
##### Description

For some reason the ditto service is always updating.

We enter [this block](https://github.com/sigma-dev/sigma/blob/master/manager/manager.go#L320) when it shouldn't 

##### Sigma Configuration

Default config 

##### Steps to reproduce

1. Start sigma
2. Add the ditto plugin 
3. Add the ditto service

##### For reference
Logs
```
INFO[2017-06-23T11:24:16-04:00] Initialized Docker Tasker                     file=sigma.go line=104
INFO[2017-06-23T11:24:16-04:00] API HTTP Server listening on 0.0.0.0:24601    file=server.go line=168
INFO[2017-06-23T11:24:19-04:00] Added Plugin: ditto                           file=server.go line=397
INFO[2017-06-23T11:24:21-04:00] Added Service: dittoSvc1                      file=server.go line=482
INFO[2017-06-23T11:24:29-04:00] Started Operation 'Deploy'                    Service=dittoSvc1
INFO[2017-06-23T11:24:37-04:00] Service Active.                               Service=dittoSvc1
INFO[2017-06-23T11:24:49-04:00] Started Operation 'Update'                    Service=dittoSvc1
INFO[2017-06-23T11:24:57-04:00] Service Active.                               Service=dittoSvc1
INFO[2017-06-23T11:25:09-04:00] Started Operation 'Update'                    Service=dittoSvc1
INFO[2017-06-23T11:25:17-04:00] Service Active.                               Service=dittoSvc1
```

PR #190, opened 2017-06-26
Add service IO equals method

helik:
Resolves: #180

### Change log

- Convert old input to correct types before comparing against new input --> but now in a nice new method!

### Notes

This is why services were always updating when they didn't only have string inputs. The old input would be a generic type (e.g. []interface{}) and therefore not be equal to the new input which was converted to the correct type (e.g. []string).

*I would want to add this to something that pulls the service out of the database, but I couldn't find a good place to do that. It seems to only matter in this one method, but it might cause problems in the future.*

### TODO
- [x] Undo everything I did on Tuesday
- [x] Undo temp fix
- [x] Create IO specific equals method in model
- [x] Write tests for new equals method


stephenrlouie:
Just looking at your comment, we should fix this for every case. 

Why isn't adding in couch a good option? We'd have some pre-processing in couchbase, but it would ensure types for all.

- [GetService](https://github.com/sigma-dev/sigma/blob/master/database/couchbase.go#L266)
- [QueryServices](https://github.com/sigma-dev/sigma/blob/master/database/couchbase.go#L456)

helik:
I didn't realize it was just a couchbase problem. That would be a good place to do it. The method I used to fix it just converts the service's input item.Actual. Is there a method for converting all of a service or should I/can I make one?

PR #197, opened 2017-06-28
Set default service.Status to Inactive

meganokeefe:
Was defaulting to empty - setting it to Inactive. 

Also added a `model.EmptyStatus` and a handler in `UpdateService` that will set an empty status to Inactive, and `log.Warn` accordingly


api/v1/convert/convert.go
```diff
@@ -111,7 +111,7 @@ func FromAPIService(o apiclient.Service) model.Service {
 		SaveDir:       o.Savedir,
 		BundleName:    o.Bundle,
 		Site:          o.Site,
-		Status:        o.Status,
+		Status:        model.ServiceStatusInactive,
```
adamschaub: Not sure if this is the place for applying a default. If the incoming API service has a status set, I reckon we should just keep whatever that status is.

manager/manager.go
```diff
@@ -346,6 +346,14 @@ func (m *manager) updateService(serviceID string) {
 				m.startServiceOperation(&service, plugin)
 			}
 		}
+	// handle a case where status has been erroneously changed to an empty string - set to Inactive
+	case model.ServiceStatusEmpty:
+		log.Warnf("Service %s has a blank Status- setting to status=Inactive", service.Name)
+		service.Status = model.ServiceStatusInactive
+		err := m.db.UpdateService(service)
```
adamschaub: Update service gets invoked when the function exits

PR #198, opened 2017-06-28
Service Operation histories 

### LINKED ISSUE #194: Store serviceOperation history
Keep track of previous serviceOperations, their outputs, whether they succeeded, etc. 
For UI purposes.

meganokeefe:
Keep the 10 most recent `serviceoperations` for a service in the database, for history/UI purposes. 
Closes #194.  

### Changelog 

- [x] `manager.removeoperation()` is now `archiveOperation()`. Rather than deleting a serviceOperation when it's completed, manager checks if we've surpassed the archive limit of 10 operations, and deletes the `x` oldest operations if needed (where `x` is the # over the limit), then 2. retains the completed serviceOperation in the DB by copying it into a new struct (see Notes) 
- [x] unit test for `archiveOperation`
- [x] API endpoint for `GetServiceOperations`, GET url would look something like: `http://127.0.0.1:24601/v1/services/helloSvc1/operations`. Returns an array of serviceOperations.

### Notes 

Right now we rely on the active serviceOperation for a service to be `<svcID>-operation`. So when we archive a competed serviceOperation, we have to give it a new ID. After some discussion, agreed to just have the new ID be a `model.GenerateUUID()`. 

### TODO 

Do we want to make the "# operations archived" number configurable?


manager/manager.go
```diff
@@ -484,16 +485,46 @@ func (m *manager) finishOperation(op model.ServiceOperation) (retErr error) {
...
 	}
-	if err := m.db.DeleteServiceOperation(op.Id); err != nil {
-		log.WithError(err).Errorf("Unable to remove service operation %s", op.Id)
+	err := m.db.DeleteServiceOperation(op.Id)
+	if err != nil {
+		log.WithError(err).Errorf("Unable to delete active serviceOperation=%s", op.Id)
+	}
+
+	// get all serviceOperations for the service corresponding to this serviceOperation
+	existingOps, err := m.db.QueryServiceOperations(model.ServiceOperation{Service: model.Service{Name: op.Service.Name}})
```
kochman: Do we want to query by `op.Service.Id` going forward?

manager/manager.go
```diff
@@ -484,16 +485,46 @@ func (m *manager) finishOperation(op model.ServiceOperation) (retErr error) {
...
+	if len(existingOps) >= archiveLimit {
+		oldestTime := time.Now()
+		var oldestID string
+		for _, o := range existingOps {
+			if o.LastUpdated.Before(oldestTime) {
+				oldestTime = o.LastUpdated
+				oldestID = o.Id
+			}
+		}
+		err = m.db.DeleteServiceOperation(oldestID)
```
kochman: This only ever deletes one service operation, so we'll have to modify it to delete multiple existing operations if we ever decrease `archiveLimit` or if it becomes user configurable.

Maybe fix it now? Potentially `for len(existingOps) >= archiveLimit` and also remove the `oldestID` operation from `existingOps`.

manager/manager.go
```diff
@@ -869,14 +900,14 @@ func (m *manager) gatherChildSitesWorker(originName, parentName string, isFirst
 		return nil, err
 	}
 	//base case - No children to recurse down or completed the cycle back to myself
-	if len(children) == 0 || originName == parentName && !isFirst{
+	if len(children) == 0 || originName == parentName && !isFirst {
```
kochman: 🙏

manager/manager.go
```diff
@@ -484,17 +485,50 @@ func (m *manager) finishOperation(op model.ServiceOperation) (retErr error) {
...
-		// when do, readd the return
-		log.WithError(err).Errorf("Unable to cleanup operation %s", op.Id)
-		// return
+		log.WithError(err).Errorf("Unable to tasker.Remove %s", op.Id)
 	}
-	if err := m.db.DeleteServiceOperation(op.Id); err != nil {
-		log.WithError(err).Errorf("Unable to remove service operation %s", op.Id)
+
+	// get all serviceOperations for the service corresponding to this serviceOperation
+	existingOps, err := m.db.QueryServiceOperations(model.ServiceOperation{Service: model.Service{Name: op.Service.Name}})
```
kochman: Do we want to query by op.Service.Id going forward?

manager/manager.go
```diff
@@ -869,14 +903,14 @@ func (m *manager) gatherChildSitesWorker(originName, parentName string, isFirst
...
-	if len(children) == 0 || originName == parentName && !isFirst{
+	if len(children) == 0 || originName == parentName && !isFirst {
 		return allChildSites, nil
 	}
 
 	// recursive case
 	for _, child := range children {
 		//originName is name of root, so keep passing it -- it should not change
-		recursiveChildren, err := m.gatherChildSitesWorker(originName ,child.Name, false)
+		recursiveChildren, err := m.gatherChildSitesWorker(originName, child.Name, false)
```
kochman: 🤙

manager/manager.go
meganokeefe: Ha. that's my linter at work.

manager/manager.go
meganokeefe: Good point! Fixing.

manager/manager.go
meganokeefe: Yeah probably a good idea.

manager/manager.go
```diff
@@ -1030,3 +1029,18 @@ func (m *manager) populateBundleInput(bs model.Service, in *model.ServiceDataIte
 func operationID(svcID string) string {
 	return fmt.Sprintf("%s-operation", svcID)
 }
+
+// implement the Sort interface for sorting serviceOpeartions by lastUpdated time
+// TODO - keep this implementation in this file?
```
kochman: I think this is okay because it's only used here. Can pull it out into `model` later if we need it somewhere else.

PR #200, opened 2017-06-28
Make Integration Tests Blaze Again

mikeynap:
420


tests/integration/manager/manager_test.go
```diff
@@ -285,13 +285,24 @@ func waitForIt() error {
...
 
+// This is a beautiful function. Gorgeous.
+// People say to me, "Mike you have the best function".
+// Trust me folks, I know nice functions when I see one.
 func ActuallyEqual(x interface{}, y interface{}) bool {
 	if reflect.DeepEqual(x, y) || x == nil && y == nil {
 		return true
 	}
+
+	if (x == nil && y != nil) || (x != nil && y == nil) {
```
808codist: at line 302 we know that at least one of x and y are not nil (from line 298), so
```
   if (x == nil || y == nil) 
'''
is sufficient, right?

mikeynap: ey nice catch! #Logic

Issue #179, opened 2017-06-22
MakeQuery Reflection instead of switch / if hell


stephenrlouie:
##### Description

Lets please use reflection on the [makeQuery function](https://github.com/sigma-dev/sigma/blob/master/database/couchbase.go#L620)

PR #214, opened 2017-07-03
Use reflection for queries in couchbase

helik:
Replace switch case in makeQuery with reflection.

Resolves: #179 

### Change log

- Removed switch case in makeQuery with a single for loop that iterates over a reflect.Value of the object to be queried
- Added a helper method to add said field to the query
- Added a test for querying bundles

### Notes

There were no tests for bundles in couchbase_test.go so I added one for querying bundles. We probably want more tests for other bundle methods. There are nice rich bundles w/plugins & services in the bundlesPopulate helper method to use for testing.

Currently, we're only querying on id and name for Bundles, do we want to be able to query with other info? For example, you could query for all bundles that have a certain service.


database/couchbase.go
```diff
@@ -618,161 +620,57 @@ func isClean(s string) bool {
...
 
-		for k, v := range obj.Config {
+func addQuery(params map[string]interface{}, query string, isFirst bool, f reflect.Value, 
+	fieldName string) (map[string]interface{}, string, bool) {
+	if f.Kind() == reflect.Ptr && !f.IsNil() {
+		f = f.Elem()
+	}
+	if (f.Kind() <= reflect.Array || f.Kind() == reflect.String || f.Kind() == reflect.Interface) && len(fmt.Sprintf("%v", f)) != 0 {
+		query, isFirst = queryHelper(isFirst, query, fieldName + "=$" + getCleanKey(fieldName))
+		params[getCleanKey(fieldName)] = f.Interface()
```
mikeynap: getCleanKey is pretty expensive [go regexes+replace are slow as hell) for an operation that doesn't matter (the param name isn't important). 

I'd recommend the "randomID" shit I did on the query branch.

mikeynap: (i.e. replace the two calls to getCleanKey() with some random string as the queryParmName)

database/couchbase.go
```diff
@@ -618,161 +620,57 @@ func isClean(s string) bool {
...
+	return strings.ToLower(toCleanReg.ReplaceAllLiteralString(s, ""))
+}
 
-		for k, v := range obj.Config {
+func addQuery(params map[string]interface{}, query string, isFirst bool, f reflect.Value, 
+	fieldName string) (map[string]interface{}, string, bool) {
+	if f.Kind() == reflect.Ptr && !f.IsNil() {
+		f = f.Elem()
+	}
+	if (f.Kind() <= reflect.Array || f.Kind() == reflect.String || f.Kind() == reflect.Interface) && len(fmt.Sprintf("%v", f)) != 0 {
```
mikeynap: this was my typo -- should be `f.Kind() < reflect.Array` (not <=)

mikeynap: Not sure if this part is right: f.Kind() == reflect.Interface -- did you find a case where that was needed?

helik: it should be <= actually -- yours was >= which should have been > so the reverse which I'm using is <=

helik: the reflect.Interface is for times when we want fields of map[string]interface{} (ex: Plugin.Config)

database/couchbase.go
helik: sure, can do

helik: actually, I should probably use your randomID since it is in the same file -- what's the best way to do that? when is the query branch going to be merged into master, it might be best to wait for that to happen?

database/couchbase.go
helik: your >= was actually correct, I was trying to reverse > because we should it was a typo which would be <= but since >= is correct the reverse is <

mikeynap: if (f.Kind() <= reflect.Array || f.Kind() == reflect.String || f.Kind() == reflect.Interface) && len(fmt.Sprintf("%v", f)) != 0 {
		query, isFirst = queryHelper(isFirst, query, fieldName + "=$" + getCleanKey(fieldName))
		params[getCleanKey(fieldName)] = f.Interface()
	}

Oh wow that snippet will work on array or maps? I didn't know that. Cool.
If that's the case, you might wanna add reflect.Slice to the list if this code will match arrays.

database/couchbase.go
mikeynap: We'll probably merge this first then I'll update my branch. Just grab it off my branch and put it here.

database/couchbase.go
helik: No, it shouldn't be handling arrays that was a typo. The next if statement handles the maps which iterates and recursively calls this method on the map element which is an interface{} so we need that f.Kind() == reflect.Interface to catch that

database/couchbase.go
helik: ok, sounds good!

PR #221, opened 2017-07-05
NomadTasker Enhancements + Fixes

meganokeefe:
#### task/nomad.go
- [x] removed "Killed" and "Killing" from the `BadNomadTaskStates` that result in `NomadTasker.Status` returning an opeartionStatus="Failed." These statuses are normal when you do a `tasker.Remove` and stop the Job. This was what was causing `TestNomadRemove` to fail, because the way we're statusing for Nomad is by checking Job allocation task states, which are not deleted right away (but instead garbage collected by nomad periodically). This caused a state in our tests where we successfully removed a Job, but are still able to Status it for a little while after. 
- [x] changed the way we ID/Name Nomad Jobs -- previously, was giving them `Service.Name`, now for lack of ambiguity reasons / debugging purposes, giving them the `operation.Id` instead. 

#### vagrant/nomad 
- [x] created a standalone nomad VM, for convenience/speed purposes. 
- [x] added instructions in `vagrant/nomad` with some troubleshooting info 

#### integration tests 
- [x] `make integration` now excludes the nomad integration tests. must run `make nomad_integration` explicitly. This command runs only the nomad integration tests, given that it starts up and shuts down a VM. Travis will run both `make integration` and `make nomad_integration`, using a local Nomad install just as before. 
- [x] removed the false positive on the nomad int. tests -- now when they fail, the Travis build fails too.
- [x] bulked out the nomad integration tests (passing and failing tests for each tasker func). 
- [x] on startup, integration tests now check where Nomad is running (local vs. VM) by pinging and checking for errors. Not the cleanest way to do this, and dependent on us not changing the nomad IPs in the Vagrantfiles -- but couldn't figure out a way to export an env variable from the nomad_deploy scripts. Ideas welcome.


task/nomad.go
```diff
@@ -96,13 +94,14 @@ func (nt *NomadTasker) Run(so *model.ServiceOperation) error {
 
 // Remove deletes this serviceOperation's Nomad Job from Nomad
 func (nt *NomadTasker) Remove(so *model.ServiceOperation) error {
-	_, _, err := nt.Wrapper.DeregisterJob(so.Service.Name, true, nil)
+	time.Sleep(time.Second * 5) //grace period for final Ship, needed for very short-running plugins
```
mikeynap: Or else what? What happens if you don't sleep?

task/nomad.go
```diff
@@ -215,7 +214,11 @@ func (nt *NomadTasker) createPluginTask(so *model.ServiceOperation) (*nomadapi.T
 	pit.EmbeddedTmpl = &(stringPrevInput)
 	templates := []*nomadapi.Template{&it, &pit}
 
-	return &nomadapi.Task{Name: "plugin", Leader: true, Driver: "docker", Env: env, Config: taskConfig, Templates: templates}, nil
+	mem := 100
+	cpu := 500
```
mikeynap: 500... arbitrary compute units?

mikeynap: :)p

task/nomad.go
mikeynap: Could it be solved by not purging? (`.DeregisterJob(id, false, nil)`)?

task/nomad.go
meganokeefe:
```
      # Specify the maximum resources required to run the job,
      # include CPU, memory, and bandwidth.
      resources {
        cpu    = 500 # MHz
        memory = 128 # MB
```

task/nomad.go
stephenrlouie: why sleep?

task/nomad.go
stephenrlouie: can you add those comments to indicate units for future people who have to edit this

tests/integration/nomad_deploy.sh
```diff
@@ -0,0 +1,27 @@
...
+  sudo chmod +x nomad
+  sudo mv nomad /usr/bin/nomad
+
+  sudo mkdir -p /etc/nomad.d
+  sudo chmod a+w /etc/nomad.d
+  echo "...Success!"
+  # Run nomad in standalone mode
+  pwd
+  sudo nomad agent -dev --config=./tests/integration/task/nomad/local_config.json &
+else
```
stephenrlouie: this does require all developers to have vagrant, should we specify this dependency

tests/integration/task/nomad/nomad_test.go
```diff
@@ -24,13 +25,30 @@ var (
 	db                 database.Database
 	nt                 *task.NomadTasker
 	useNomadUrl        = "127.0.0.1:4646"
+	vmNomadDefault     = "172.17.7.100:4646"
```
stephenrlouie: 172.17.7 will fall in my default container IP space of 172.17.0.0./16. Can we move it to `172.16.`

tests/integration/task/nomad/nomad_test.go
```diff
@@ -43,8 +61,9 @@ func TestMain(m *testing.M) {
 	fmt.Printf("\nInstantiated nomadtasker: %+v", nt)
 
 	// run all tests
-	m.Run()
```
stephenrlouie: oops

tests/integration/task/nomad/nomad_test.go
```diff
@@ -24,13 +25,30 @@ var (
...
+			fmt.Println("Using VM Nomad client instead of localhost")
+			useNomadUrl = vmNomadDefault
+		}
+	} else {
+		fmt.Println("Using localhost for Nomad client")
+	}
+
+	// init nulldb for API
 	db = database.NewNullDB(false)
 	go startAPI(db)
```
stephenrlouie: we should have a way to stop this go routine. Once the parent process ends it should close, so not a huge deal.

tests/integration/task/nomad/nomad_test.go
```diff
@@ -107,61 +126,163 @@ func TestCreateNomadJob(t *testing.T) {
...
+			false,
+		},
+	}
+	for i, test := range tests {
+		err := nt.Run(test.input)
+		if test.expectPass {
+			if err != nil {
+				t.Errorf("Test %d expected no err, got: %v", i+1, err)
+			} else {
+				t.Logf("Test %d successfully passed without errors", i+1)
```
stephenrlouie: just to shorten tests, not sure if we want that each passed. I think silence could be implied as success?

tests/integration/task/nomad/nomad_test.go
```diff
@@ -107,61 +126,163 @@ func TestCreateNomadJob(t *testing.T) {
...
+			}
+			time.Sleep(time.Second * 3)
+		}
+		err := nt.Remove(test.op)
+		if test.expectPass {
+			if err != nil {
+				t.Errorf("Test %d expected no Remove err, got: %v", i+1, err)
+			} else {
+				// Check status, should be Finished
+				time.Sleep(time.Second * 2)
```
stephenrlouie: these sleeps scare me. Is there no way to call nomad to ask if something is done?

tests/integration/nomad_deploy.sh
mikeynap: Developers who wanna run this specific integration test

tests/integration/task/nomad/nomad_test.go
mikeynap: don't think this is a priority

tests/integration/nomad_deploy.sh
stephenrlouie: yea, we also have a docker dependency, not sure if we need HAVE to document. maybe for someone completely new to the project

Makefile
```diff
@@ -18,7 +19,12 @@ integration:
 	./tests/integration/deploy.sh
 	go test -v  --ldflags '${EXTLDFLAGS}' $(PACKAGES_FOR_INTEGRATION)
 
+nomad_integration:
+	./tests/integration/nomad_deploy.sh
+	go test -v --ldflags '${EXTLDFLAGS}' $(PACKAGES_FOR_NOMAD_INTEGRATION)
+
 clean_integration:
+	./tests/integration/nomad_stop.sh
```
stephenrlouie: if nomad has its own set up, it should have its own break down

meganokeefe: okey doke

tests/integration/task/nomad/nomad_test.go
meganokeefe: I know...

tests/integration/nomad_deploy.sh
```diff
@@ -0,0 +1,27 @@
+# Start nomad locally if we're on Linux; use VM otherwise
+if [ "$(expr substr $(uname -s) 1 5)" = "Linux" ]; then
```
stephenrlouie: I see `expr: syntax error`

vagrant/nomad/multi-node/Vagrantfile
```diff
@@ -12,6 +12,7 @@ cd /tmp/
 curl -sSL https://releases.hashicorp.com/nomad/0.5.6/nomad_0.5.6_linux_amd64.zip -o nomad.zip
```
stephenrlouie: Can we make a base image that has nomad and docker installed? Will save us time over the long term, since we all bite the bullet downloading the big image once. Then just have it. instead of installing every time.

mikeynap: ^^^

tests/integration/nomad_deploy.sh
mikeynap: NO. THEY CAN FIGURE IT OUT THE HARD WAY

tests/integration/nomad_deploy.sh
mikeynap: I'm not sure if the quotes around $( is kosher

vagrant/nomad/multi-node/Vagrantfile
mikeynap: There must be an image with dockeR and nomad installed somewhere right?

stephenrlouie:
```
=== RUN   TestNomadRemove
--- FAIL: TestNomadRemove (1.02s)
	nomad_test.go:205: Remove unsuccessful- got error on nomadtasker.Status=No allocations found for job=8387b53b-7c67-4241-8008-6b233a1381fe
=== RUN   TestNomadStatus
--- FAIL: TestNomadStatus (0.02s)
	nomad_test.go:270: No allocations found for job=e1e56f7a-a407-4aa0-86fd-5689045b2fcd
```

tests/ci.sh
```diff
@@ -6,4 +6,5 @@ function clean {
 }
 trap clean EXIT
 make test
+make nomad_integration
```
stephenrlouie: lets add the nomad_clean to the clean call

tests/integration/task/nomad/nomad_test.go
```diff
@@ -179,49 +173,60 @@ func TestNomadRemove(t *testing.T) {
...
 				t.Error(err)
 			}
-			time.Sleep(time.Second * 3)
 		}
-		err := nt.Remove(test.op)
+		removeErr := nt.Remove(test.op)
+		tries := 0
+		//give the job a chance to run before removing
+		if removeErr != nil {
+			for tries < maxRetry && strings.Contains(strings.ToLower(removeErr.Error()), "no allocations found") {
```
stephenrlouie: huh retry here?

tests/integration/nomad_deploy.sh
meganokeefe: Removing the quotes gets me `./tests/integration/nomad_deploy.sh: line 2: [: =: unary operator expected` 

I cannot bash

tests/integration/task/nomad/nomad_test.go
```diff
@@ -26,13 +28,31 @@ var (
 	db                 database.Database
 	nt                 *task.NomadTasker
 	useNomadUrl        = "127.0.0.1:4646"
+	vmNomadDefault     = "172.16.7.100:4646"
 	useAPIUrl          = "127.0.0.1:24608"
 	useTimeoutDuration = "1m"
+	maxRetry           = 10
```
stephenrlouie: running a bunch of openshift on my local host was showing that 10 might not be enough. can we up it to like 60 so a minute.

PR #223, opened 2017-07-06
Rekick

vgunning:
### PR checklist

### Change log

Adds a `ResetService` function to `api/v1/server/server.go`
Adds endpoint to be generated in swagger documentation
Adds a route to the `api/v1/server/routers.go`

Reset deletes all the `Failures` from a service and changes the status to `model.ServiceStatusInactive` instead of failed.


mikeynap:
1. Have you successfully run a (physical) test of this? I.e. service is in a failed state, you call /reset, it tries again?
2. Are we sure this is safe for services that are currently in "Active" or "Running" or whatever state? What if there is a container running for a service and this is called? What will happen it to? What will happen to the old service operation?

vgunning:
@mikeynap 
1. I've worked successfully reset failed services (i.e. not setting the docker save path to something it can write to and watching services fail/timeout. reset does in fact let them fail/timeout all over again)
2. We'd have to decide whether we want resetting a service to trump an `Active` service. Or if we're only going to allow a reset on services in the failed state. If it's the latter, that's an easy check to add.

mikeynap:
Reset seems to imply that you're gonna stop/start a service, which isn't really the case.

I'd feel way better if the code as is now had a check to ensure it was in a failed state before updating the service.

It might be worth it to have support for restarting the service, but I think that might be separate bag of worms (that requires deeper level changes).

vgunning:
Yea, I think that's out of the scope of this PRs intent. Which was to allow a user to reset a service that's entered a failed state.
@adamschaub thoughts?

mikeynap:
StatusMethodNotAllowed (http 405) is not the right code to return. That exclusively means "You posted, but I don't allow posts".

Return a BadRequest and make the error slightly more descriptive, like "Reset not allowed on active services" or something.

Issue #222, opened 2017-07-06
Bad Warnings...


mikeynap:
STDOUT is being blasted with warnings about: 
```Couldn't convert Service I/O. Got <nil>. Expected type whatever```
and
```2017/07/06 13:37:37 http: multiple response.WriteHeader calls```

PR #224, opened 2017-07-06
Return empty list of provided type

adamschaub:
Addresses "expected type" error in #222 

For any pointer-type values (maps, slices), a base value of `nil` should be converted to an an initialized (but empty, possibly nil) equivalent (via reflection). 

Make sure it works for:
- [x] Maps/objects
- [x] Slices of all types


model/convert.go
```diff
@@ -14,10 +13,12 @@ func ConvertIO(a interface{}, t string, plug Plugin) (interface{}, bool) {
 		typ := reflect.ValueOf(a)
 		var l []interface{}
 		if typ.Kind() != reflect.Slice {
-			if v, ok := ConvertIO(a, out, plug); ok {
-				l = append(l, v)
-			} else {
-				return nil, false
+			if a != nil {
```
helik: So what happens if a == nil? It returns a, false in line 42?

model/convert_test.go
```diff
@@ -355,12 +374,11 @@ func TestConvertPluginIO(t *testing.T) {
 			can: true,
 		},
 		// Ones that don't work
```
helik: Should this comment be moved since the next test does work?

model/convert.go
adamschaub: it returns `final, true` if it correctly passes `toListOfType`

helik: ah gotcha, didn't see that

PR #244, opened 2017-07-12
Changing Service Enabled to a Disabled flag

### LINKED ISSUE #233: Returned Query Objects don't Contain Enabled False field
##### Description

When querying for objects, the returned objects don't contain the `enabled:False` field. Unsure if this is true for all fields that are false.

##### Sigma Configuration

Insert service with an `enabled: False` field
Query `curl -X GET -H  "Content-Type: application/json" http://127.0.0.1:24601/v1/services?enabled=false`
Inspect objects and see a lack of `enabled` field

https://stackoverflow.com/questions/37756236/json-golang-boolean-omitempty
Describes the situation nicely...

##### Suggest a Fix
We should change our field labeling so that we can assume the default value when a field is omitted.
`Enabled` -> `Disabled`

vgunning:
### PR checklist

### Change log

Changes the swagger docs to have `Disabled` with a default to `False` instead of `Enabled`
Updates the swagger html
Updates `convert.go` to use an `IsDisabled()` check/function instead of `IsEnabled()`

### Notes

Resolves #233 

### TODO

Vendoring for `mu`?
Verify everything in UI was changed correctly.


api/v1/gen/v1.html
```diff
@@ -229,6 +229,7 @@
   <li><a href="#getUser"><code><span class="http-method">get</span> /secrets/users/{user-name}</code></a></li>
   <li><a href="#getUsers"><code><span class="http-method">get</span> /secrets/users</code></a></li>
   <li><a href="#loginWithStoredUsernameAndPassword"><code><span class="http-method">get</span> /login</code></a></li>
+  <li><a href="#resetService"><code><span class="http-method">get</span> /services/{service-name}/reset</code></a></li>
```
meganokeefe: where did `resetService` come from?

vgunning: I didn't run the swagger html when this PR got merged https://github.com/sigma-dev/sigma/pull/223

PR #252, opened 2017-07-13
Sigma REBIRTH Proposal

adamschaub:
See doc


docs/pluginv2.md
```diff
@@ -0,0 +1,99 @@
...
+maintainer: Adam Schaub <adamscha@cisco.com>
+version: 0.0.1 # must be numeric-dot notation so we can declare accepted ranges
+vars:
+  var_name:
+    description: |
+      A (possibly long) description of what the var provides for the plugin
+    summary: Shorter description
+    type: list/string # types as before
+    default: [] # The value used if not explicitly set by the service.
+    required: true # true iff this fact must be specified. Satisfied if a default is supplied, so specifying a default makes this unnecessary
```
adamschaub: OR use `optional: true` for fields that may be set by the plugin. Depends if we anticipate the average variable will be required or optional. Then pick the less frequent one.

docs/pluginv2.md
```diff
@@ -0,0 +1,99 @@
...
+              # to other plugins
+  # Versions conditions should be listed in ascending order
+  # Conditions that represent discontinuous ranges are not allowed (i.e. <1, >=2)
+  # != and == "exclude" and "include" versions explicitly and don't obey discontinuity (i.e. >1, ==0.1, !=1.1 is allowed)
+  - plugin-one:
+      version: [>version, >=version, <version, <=version, !=version, ==version]
+      deps: [requires]
+      # Can specify multiple plugins to satisfy this requirement
+      # I envision needing info about multiple services that satisfy the same plugin
+      allows_multiple: true
```
mcaulfield: I think we need to be able to specify something like `?, +, *` where `?` is 0 or 1 instances, `+` is 1 or more, and `*` is 0 or more. Could collapse deps and allows_multiple into a single field that covers this? or even use the actual `?, +, *` notation. We could get super terse and do something like this:
```
requirements:
    - plugin-one* >= 3.4
    - plugin-two+ == 1.0
    - plugin-foo?
```
Maybe that's going too far.

docs/pluginv2.md
```diff
@@ -0,0 +1,99 @@
...
+  another_var:
+    description: |
+      This var may be supplied after-the-fact, or is not needed. It is,
+      however, guaranteed to exist for other plugins that use this plugin
+    summary: A place holder
+    type: object # still can use type. If no specified, it's laissez faire
+# Requires lists plugins and allowed versions that must be specified for a service
+# to be deployed. If one or more required services are not active, then this plugin
+# cannot be expected to operate properly.
+requirements: # TODO: Suggest a better name? Expresses how this plugin interrelates
```
mcaulfield: I'm good with `requirements` or `reqs` for short?

docs/pluginv2.md
```diff
@@ -0,0 +1,99 @@
...
+name: a-unique-name
+plugin: my-plugin-in
+plugin_version: 0.0.1
+site: BXB-central
+# List services to provide this service's plugin. Must at minimum satisfy the
+# plugin's requires
+include:
+  - plugin-one: [plugin-one-svc]
+  - plugin-two: [plugin-two-svc]
+requires:
```
mcaulfield: I don't get why `include` and `requires` are separate sections. Why not have a single section that lists all the services that this service depends on?

docs/pluginv2.md
```diff
@@ -0,0 +1,99 @@
...
+  # Inherit deps for each service based on plugin
+  # Add additional requirements that may not be specified in plugin
+vars:
+  var_name:
+    value: ["list", "of", "strings"]
+    secret: true
+```
```
mcaulfield: I think we can eliminate the bundle concept. Instead we can allow for bulk service creation. We might also allow for the creation of a stub service (one that doesn't allow the container to be run) and maybe a stub plugin (which doesnt even specify a container). These are special cases but just variants of the plugin/service construct. No need for any additional api endpoints and no need to implement them immediately. Hopefully this further simplifies Sigma.

docs/pluginv2.md
```diff
@@ -0,0 +1,99 @@
...
+allow a consumer to better understand how the plugin works from inspecting
+the API document alone.
+
+Second, it expressly declares associated plugins (and their versions) that need
+to be present to configure this service. The vars expressed in these plugins
+will then be made available to the plugin for use in operations.
+
+```
+---
+# Consider tagging it with a plugin API version
```
adamschaub: So do we want this change to make v2? Or still use v1 for now, considering it's sort of alpha atm

docs/pluginv2.md
adamschaub: I think so. The default value might stand as a further example of what the value should be. And optional vars might have defaults as well, so I'm not sure if default==required.

docs/pluginv2.md
adamschaub: `reqs` works for me. It was previously `richness`

docs/pluginv2.md
adamschaub: It's good to be short, but I'm not sure the value. Given time writing plugin documents is a one-time cost, and I'm not convinced it makes it more readable, I'd give this a luke-warm sure

docs/pluginv2.md
adamschaub: `include` tells sigma to make the vars from this the listed services available to the plugin. You would only include a from plugins that your plugin writer has specifically made provisions for.

`requires` gives contextual information about how this service responds to changes in other services, i.e. blocking for a required service to become ready, potential down the line for triggering targeted updates, etc.

docs/pluginv2.md
adamschaub: I'm guessing that would be fine. I don't see why someone couldn't write a collection of services that would be tweaked and reused across multiple deployments in the same way. Bundles seem like an optimization until we have a deployment on hand that really pushes the ease of use of services/plugins alone.

docs/pluginv2.md
stephenrlouie: The way I would understand this is optional would never have a default associated. Maybe we only have optional: true, or you define a default which means it will always have a value (`optional: false`), it's either one you declared or the default provided.

docs/pluginv2.md
stephenrlouie: Not understanding why these fields are in the service definition, is their presence in the plugin definition not sufficient. Do we have use cases where you won't know this information at plugin time

adamschaub: Some plugins are optional and don't have to be included. At a minimum, you must include services that fulfill your required plugins. You can add others that may be optional.

docs/pluginv2.md
kochman: I think that `requirements` is clearer than `reqs`.

docs/pluginv2.md
mcaulfield: I'm saying that if a field has a default value, then we can imply that the field is optional (i.e. need not be specified by the user). It sounds like @adamschaub you are saying that even for a required field you might use the concept of a default as an example?

docs/pluginv2.md
mcaulfield: Everything is v1 for now.

docs/pluginv2.md
mcaulfield: I just don't like `allows_multiple` - something like `quantity` would be better.

docs/pluginv2.md
mcaulfield: I think for now we only have a need for the `include` concept. However I'd rather use the same naming (`reqs`) that we use in the plugin definition.

docs/pluginv2.md
adamschaub: +1 agreed

docs/pluginv2.md
vgunning: Does this complicate that there could be for instance multiple plugins that provide the same kind of output, and any of those plugins would satisfy the requirement?

docs/pluginv2.md
adamschaub: I'll make sure it matches

docs/pluginv2.md
adamschaub: I'll find a way to simplify this. I'll reword the proposal to be a list or plugin -> string map containing version conditions. 

@helik Also brought up a point about allowing range discontinuities, which I'll consider. For simplicity in the moment though, I'm going to aim just at contiguous (all logical ANDs) aside from `!=` and `==` conditions.

docs/pluginv2.md
mikeynap: requires: is another possibility

mikeynap: or needs:

docs/pluginv2.md
mikeynap: I can't think of a use case for discontinuous ranges lol.

docs/pluginv2.md
mikeynap: If we don't need to "wire" up the inputs/outputs, there is less of a need for a bundle.

docs/pluginv2.md
```diff
@@ -25,46 +25,66 @@ will then be made available to the plugin for use in operations.
 
 ```
 ---
-# Consider tagging it with a plugin API version
-name: my-plugin-name
-description: In depth details about how we go about doing the thing
-summary: This does thing
-maintainer: Adam Schaub <adamscha@cisco.com>
-version: 0.0.1 # must be numeric-dot notation so we can declare accepted ranges
+sigma_info:
```
mikeynap: Why not keep it top level? 
At least name it something like "header", sigma_info seems pretty bleh

docs/pluginv2.md
```diff
@@ -25,46 +25,66 @@ will then be made available to the plugin for use in operations.
...
+    # Later: provide another namespace/owner/something to distinguish two
+    # plugins with the same name.
+  - name: plugin-two
+    description: Does the two things. Uses variables xyz to accomplish w
+    # No version/empty specified means we don't care what version
+    # version:
+# Optional has the same format as requirements, except no plugins here are
+# required for the plugin to operate.
+# Plugins here may have 0 or more instances included.
+optional:
```
mikeynap: What is the purpose of "optional" plugins? Is it like "The functionality of this plugin MAY be extended In the presence of optional plugins"?

docs/pluginv2.md
```diff
@@ -81,17 +101,25 @@ purposes. At a minimum, included services must satisfy the required plugins.
...
-requires:
-  # Inherit deps for each service based on plugin
-  # Add additional requirements that may not be specified in plugin
+# plugin's requirements.
+# Each included service is matched automatically to any declared required or
+# optional plugins based on plugin name and version constraints.
+# Services here must be in an `Active` state for this service to be considered
+# in good health/able to deploy.
+requirements:
+  - plugin-one-svc
```
mikeynap: Is it required to list the service-names here? 
Would it be beneficial to allow sigma to "discover" a service using the "required" plugins? That is:

Assuming plugin `a` requires plugins `b`,`c` and there exists `b-svc` and `c-svc` already in sigma: 
Could a service `a-svc` automatically use `b-svc`/`c-svc` if they were not explicitly named as requirements in the `a-svc` definition?

docs/pluginv2.md
```diff
@@ -0,0 +1,127 @@
...
+allow a consumer to better understand how the plugin works from inspecting
+the API document alone.
+
+Second, it expressly declares associated plugins (and their versions) that need
+to be present to configure this service. The vars expressed in these plugins
+will then be made available to the plugin for use in operations.
+
+```
+---
+sigma_info:
```
mikeynap: Why not keep it top level?
At least name it something like "header"...
 `sigma_info` seems pretty bleh

adamschaub: It looked better to be separated, given it's all information used by sigma for organization. Could do header or w/e. Very conspicuous details can be changed quickly in the final hour

docs/pluginv2.md
adamschaub: Maybe, but that's sort of out of scope for the initial rework. Maybe in the future we allow selection/discovery based on certain attributes. For now direct seems the least convoluted

docs/pluginv2.md
adamschaub: Correct. For instance, a prometheus plugin has a bunch of optional exporters and alert managers that you might install separately. So I might write different plugins to install node_exporters in specific ways. And include them optionally here so the prometheus server knows how to find them.

docs/pluginv2.md
adamschaub: I agree

adamschaub: Especially doubly if we do something like discovery that you mentioned earlier on.

docs/pluginv2.md
adamschaub: I briefly tooled around with a labels/tags concepts sort of like what k8s does to select pods for replicasets/deployments and so on. Give the user the ability to specify arbitrary tags to lump services together.

docs/pluginv2.md
```diff
@@ -0,0 +1,127 @@
...
+# Each included service is matched automatically to any declared required or
+# optional plugins based on plugin name and version constraints.
+# Services here must be in an `Active` state for this service to be considered
+# in good health/able to deploy.
+requirements:
+  - plugin-one-svc
+  - plugin-another-one-svc
+  - plugin-two-svc
+# Optional config
+optional: [plugin-foo-svc]
```
helik: if optional plugins have the same format in the plugin, shouldn't it have the same format here?

adamschaub: Same diff as far as YAML is concerned. Both are lists of strings.

helik: Gotcha, then things look good to me

Issue #248, opened 2017-07-12
Cycle Warnings?


stephenrlouie:
##### Description

We should block service Cycles from occurring when we `AddService` or `UpdateService`

do we also have to block `AddSite` and `UpdateSite`, functionally not a bug. But a weird pattern that could lead to inconsistent fetching of children. Since its a map, if I get around the cycle to my generation, my siblings could show up first when we iterate through the map some times, or other times I could go first. Making the children list either include my siblings children or not.

stephenrlouie:
closed by #257

PR #257, opened 2017-07-13
Cycle warning on sites

stephenrlouie:
### Change log

- Now prints a warning when we detect a cycle for sites on manager. `go test -v -run TestGatherChildSites`
- The logic for checking test results was not correct. Tests would pass if the received output was just a subset of the expected output
- The mock.DB was not loaded properly to return the proper children of parents. 

### Note

Does not address #248 since that calls for API changes, this simply prints it in the manager and could be considered a fix that is "good enough" given the time constraints.


adamschaub:
I don't like it

Issue #208, opened 2017-06-29
DB object reference consistency


stephenrlouie:
##### Description

If two objects reference each other, say the `Service.Site` field. Should that value be the `Name` or `ID`. It shouldn't be a problem if you `GetSite` since we try both name and ID. 

But if I want to find all services that refer to my Site, I will have to query on `Service.Site: <ID>` and `Service.Site: <Name>` to be sure I catch everything since we are not strict on what will be stored in those fields.

mcaulfield:
tldr: Switch from using names for references to using IDs everywhere consistently.

PR #268, opened 2017-07-20
Rebirth

adamschaub:
### Change log

Update API elements and service operation input/output handling according to rebirth proposal doc.

- [x] Update model definitions
- [x] Update API definitions
- [x] Update model/API conversion functions
- [x] Update API server
- [x] Update manager
- [x] Update database
- [x] Update ALL TESTS


api/v1/convert/convert.go
```diff
@@ -79,144 +62,159 @@ func FromModelServiceOperationLogs(o model.ServiceOperationLogs) apiclient.Servi
...
+			Name:        req.Name,
+			Description: req.Description,
+			Version:     req.Version,
+		})
+	}
+	return
+}
+
+func ToModelPluginTypes(o map[string]apiclient.TypeEntry) map[string]model.Type {
+	if len(o) == 0 {
```
stephenrlouie: why is this the only one with a `len` check? is the zero value not sufficient?

api/v1/convert/convert.go
```diff
@@ -290,3 +290,13 @@ func FromAuthUser(o user.User) apiclient.User {
 		Role:  int32(o.Role()),
 	}
 }
+
+func newMetaImpl(id, name string) model.MetaImpl {
+	mi := model.MetaImpl{
```
stephenrlouie: why do many of these objects also have a `Name` field?

api/v1/server/server.go
```diff
@@ -423,20 +424,30 @@ func (s *Server) AddService(w http.ResponseWriter, r *http.Request) {
...
-				return
-			} else if len(plugins) == 0 {
-				http.Error(w, fmt.Sprintf("No matching plugin"), http.StatusBadRequest)
-				return
-			}
-			plugin = plugins[0]
-		} else {
-			http.Error(w, fmt.Sprintf("Unable to verify plugin: %s", err.Error()), http.StatusInternalServerError)
+	//TODO: Separate them
+	// Split plugin version/name
```
stephenrlouie: so the specified format will be `PluginName:Version`?

api/v1/server/server.go
```diff
@@ -423,20 +424,30 @@ func (s *Server) AddService(w http.ResponseWriter, r *http.Request) {
...
+	pluginPts := strings.Split(apiObj.Plugin, ":")
+	if len(pluginPts) != 2 {
+		http.Error(w, "Plugin name formatted incorrectly", http.StatusBadRequest)
```
stephenrlouie: we can be more specific, to guide the user.

api/v1/server/server.go
```diff
@@ -423,20 +424,30 @@ func (s *Server) AddService(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	if plugins, err := s.db.QueryPlugins(model.Plugin{Name: pluginPts[0], Version: pluginPts[1]}, nil); err == nil {
```
stephenrlouie: We should just monitor that this is updated to reflect. https://github.com/sigma-dev/sigma/issues/208

model/model.go
```diff
@@ -9,79 +9,23 @@ Jobs: JSON-defined tasks able to run a container job(s) on one or more Sites or
...
-
-// Base Types
 const (
-	TypeObject = "object"
-	TypeInt    = "int"
-	TypeFloat  = "float"
-	TypeBool   = "bool"
-	TypeString = "string"
-	TypeList   = "list" // all lists must be homogeneous
+	ID         = "Id" //The constant to search on for ID
```
stephenrlouie: Do we need both this  and [metadata.go](https://github.com/sigma-dev/sigma/blob/rebirth/model/metadata.go#L15)

model/vet.go
```diff
@@ -1,62 +1,34 @@
...
 // the declared inputs and outputs of the service
 // Requires plugin for any type definitions
 func ValidateServiceIO(svc Service, plug Plugin) (errs []error) {
-	for name, input := range svc.Input {
-		// If an input declares a value, type must be specified
-		if _, ok := ConvertIO(input.Value, input.Type, plug); input.Value != nil && !ok {
-			errs = append(errs, fmt.Errorf("Input %s default value doesn't match declared type", name))
+	for name, v := range svc.Vars {
+		// Var must exist in plugin
+		plugVar, ok := plug.Vars[name]
```
stephenrlouie: use id internally? https://github.com/sigma-dev/sigma/issues/208

api/v1/server/server.go
```diff
@@ -704,17 +693,17 @@ func (s *Server) DeletePlugin(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
-	if err := s.db.DeletePlugin(id); err != nil {
-		http.Error(w, err.Error(), http.StatusNotFound)
-		return
-	}
 	ss, _ := s.db.QueryServices(model.Service{}, getFilters(r))
 	for _, p := range ss {
-		if id == p.PluginName {
+		if id == p.Plugin.GetName() {
```
stephenrlouie: I think its fine to take in the name, we just have to convert it to an ID internally.

api/v1/server/server_test.go
```diff
@@ -215,16 +233,16 @@ func TestValidateService(t *testing.T) {
 		service   apiclient.Service
 		errExpect error
 	}{
-		{scrap, ErrDuplicateName},
+		//{scrap, ErrDuplicateName},
```

api/v1/server/server_test.go
```diff
@@ -824,6 +733,32 @@ func TestResetService(t *testing.T) {
 	}
 }
 
+// TODO - this is broken
```
stephenrlouie: lets make an issue

api/v1/server/server_test.go
```diff
@@ -834,301 +769,166 @@ func marshalOrDie(t *testing.T, v interface{}) string {
...
 }
 
-func serviceEQ(pl1 apiclient.Service, pl2 apiclient.Service) bool {
+func apiServicesEqual(pl1 apiclient.Service, pl2 apiclient.Service) bool {
 	pl1.Id = ""
 	pl2.Id = ""
 	return reflect.DeepEqual(pl1, pl2)
 }
 
+// Ignore metadata (TODO - better way to do this? getting around empty vs. nil metadata maps)
```
stephenrlouie: I guess we write our own equals?

database/couchbase.go
```diff
@@ -354,91 +318,81 @@ func (cb *Couchbase) GetSite(id string) (model.Site, error) {
 	return site, replaceErr(err)
 }
 
-func (cb *Couchbase) InsertPlugin(m model.Plugin) (string, error) {
-	m.Set("Type", PluginType)
-	m.Id = model.GenerateUUID()
-	if err := cb.insertId(m.Id, &m, 0); err != nil {
+func (cb *Couchbase) insert(m model.Metadataer, objType string) (string, error) {
+	m.Set("Type", objType)
+	if len(m.GetID()) == 0 {
```
stephenrlouie: should we error if someone tries to have an ID already assigned on insert?

database/couchbase.go
```diff
@@ -354,91 +318,81 @@ func (cb *Couchbase) GetSite(id string) (model.Site, error) {
...
 func (cb *Couchbase) InsertService(m model.Service) (string, error) {
-	m.Set("Type", ServiceType)
-	m.Id = model.GenerateUUID()
-	m.LastUpdated = time.Now()
-	if err := cb.insertId(m.Id, &m, 0); err != nil {
-		return "", replaceErr(err)
-	}
-
-	return m.Id, nil
+	m.SetTime(model.SigmaMetaTimeUpdated, time.Now())
```
stephenrlouie: do we need to handle [CreatedTime](https://github.com/sigma-dev/sigma/blob/rebirth/model/metadata.go#L18) too?

database/couchbase.go
```diff
@@ -354,91 +318,81 @@ func (cb *Couchbase) GetSite(id string) (model.Site, error) {
...
+// UPDATE
+func (cb *Couchbase) update(up model.Metadataer, objType string) error {
+	if len(up.GetID()) == 0 {
+		return ErrEmptyID
 	}
-	if err := cb.insertId(m.Id, &m, 0); err != nil {
-		return "", replaceErr(err)
+	var err error
+	var cas gocb.Cas
+	if casInterface, ok := up.Get("cas"); ok && casInterface != nil {
```
stephenrlouie: why are we storing cas?

database/couchbase.go
```diff
@@ -354,91 +318,81 @@ func (cb *Couchbase) GetSite(id string) (model.Site, error) {
...
 
-	return m.Id, nil
-}
-
-func (cb *Couchbase) InsertSite(m model.Site) (string, error) {
-	m.Set("Type", SiteType)
-	m.Id = model.GenerateUUID()
-	if err := cb.insertId(m.Id, &m, 0); err != nil {
-		return "", err
+	up.Delete("cas")
```
stephenrlouie: we might want to just put these across the board anytime an entry is made into the DB

database/couchbase.go
stephenrlouie: I'm pretty sure we can get rid of this block

database/couchbase.go
stephenrlouie: lets assign a `cas` const

database/nulldb.go
```diff
@@ -12,9 +12,7 @@ import (
 )
 
 const (
-	idKey   = "__nulldb__.id"
-	nameKey = "__nulldb__.name"
-	casKey  = "__nulldb__.cas"
+	casKey = "__nulldb__.cas"
```
stephenrlouie: we can share a `cas` const once its defined

database/nulldb.go
```diff
@@ -244,32 +240,28 @@ func (n *NullDB) insertWithName(obj model.Metadataer, objType, objID, objName st
...
 
 func (n *NullDB) InsertPlugin(o model.Plugin) (string, error) {
-	return n.insertWithName(&o, PluginType, o.Id, o.Name)
+	return n.insertWithName(&o, PluginType, o.GetID(), o.Name)
 }
 
 func (n *NullDB) InsertService(o model.Service) (string, error) {
-	o.LastUpdated = time.Now()
-	return n.insertWithName(&o, ServiceType, o.Id, o.Name)
+	o.SetTime(model.SigmaMetaTimeUpdated, time.Now())
```
stephenrlouie: again do we need created_time too

manager/manager.go
```diff
@@ -351,104 +299,276 @@ func (m *manager) updateService(serviceID string) {
...
+				Generated: v.Generated,
+			}
+		}
+		if sv.Value == nil && v.Default != nil {
+			// Set default if no value
+			sv.Value = v.Default
+		}
+		// Copy over type regardless
+		sv.Type = v.Type
+		service.Vars[name] = sv
```
stephenrlouie: is `name` unique enough or should we `name:version`. We could also convert to IDs internally?

manager/manager.go
```diff
@@ -351,104 +299,276 @@ func (m *manager) updateService(serviceID string) {
...
-		log.WithError(err).Errorf("Could not remove service %s", service.Name)
+	if err := m.db.DeleteService(service.GetID()); err != nil {
+		log.WithError(err).Errorf("Could not remove service %s", service.GetName())
 	} else {
-		log.Infof("Removed %s", service.Name)
+		log.Infof("Removed %s", service.GetName())
+	}
+}
+
+func (m *manager) getReqServices(service model.Service, plugin model.Plugin) (out []model.Service, err error) {
```
stephenrlouie: mark

manager/manager.go
```diff
@@ -351,104 +299,276 @@ func (m *manager) updateService(serviceID string) {
...
+// Call the returned function to return the service and vars map to their
+// original state
+func tempRemoveGenVars(service model.Service, vars model.ServiceVars) func() {
+	// vars[service.Plugin][0] always exists
+	restoreVars := vars[service.Plugin.GetName()][0]
+	vars[service.Plugin.GetName()] = vars[service.Plugin.GetName()][1:]
+
+	var restorePreviousVars map[string]model.ServiceVar
+	if service.PreviousVars != nil {
+		restorePreviousVars = service.PreviousVars[service.Plugin.GetName()][0]
```
stephenrlouie: is [0] guaranteed to be there?

manager/manager.go
```diff
@@ -351,104 +299,276 @@ func (m *manager) updateService(serviceID string) {
...
+	return active
+}
+
+// Temporarily removes the service's generated outputs for comparision
+// Call the returned function to return the service and vars map to their
+// original state
+func tempRemoveGenVars(service model.Service, vars model.ServiceVars) func() {
+	// vars[service.Plugin][0] always exists
+	restoreVars := vars[service.Plugin.GetName()][0]
+	vars[service.Plugin.GetName()] = vars[service.Plugin.GetName()][1:]
```
stephenrlouie: what happens if only one exists

stephenrlouie: I'm missing something here, why are we pulling off the first entry

api/v1/convert/convert.go
adamschaub: You mean `MetaName`?

api/v1/server/server.go
adamschaub: / was used here as `and`. "Split plugin version [and] name". But sure.

api/v1/server/server.go
adamschaub: I don't think we know too much about exactly why it is wrong, and I don't think it's this API's jobs to give specifics on why something isn't appropriate. Should be sufficient to say aspect failed the test, and defer to documentation for the user to figure out.

api/v1/server/server.go
adamschaub: agreed

api/v1/convert/convert.go
adamschaub: Removed

database/couchbase.go
adamschaub: I don't see why we would be picky. We also use it for a lot of testing.

database/couchbase.go
adamschaub: Could. Not sure how important it si.

database/couchbase.go
adamschaub: How do we get the value of `var cas gocb.Cas` then?

database/couchbase.go
adamschaub: Sure to the "cas" const. I'm not sure what you mean by "just put these across the board"

database/nulldb.go
adamschaub: Don't know.

model/model.go
adamschaub: Removed

model/vet.go
adamschaub: for variables in a plugin?

manager/manager.go
```diff
@@ -351,104 +299,285 @@ func (m *manager) updateService(serviceID string) {
...
+
+func (m *manager) varsChanged(service model.Service, vars model.ServiceVars) bool {
+	// Need to remove this service's generated vars from both vars and PreviousVars
+	// before comparing. Generated vars can trigger unintentional updates
+	restore := tempRemoveGenVars(service, vars)
+	defer restore()
+
+	a, _ := json.Marshal(&(service.PreviousVars))
+	b, _ := json.Marshal(&vars)
+	return !bytes.Equal(a, b)
```
mikeynap: idk if this works. I think bytes.Equal will ALWAYS be false because you're unmarshalling two maps. The order of the maps will always be different, so it will always report a change... right?

manager/manager.go
mikeynap: ^^^^. are you sure vars will always contain[service.Plugin.GetName()]? and will have two entires?

database/couchbase.go
stephenrlouie: we have to do `cb.Get`

stephenrlouie: its good, actually

database/couchbase.go
stephenrlouie: the DB should be responsible for stripping the cas anytime people update or insert that code should `Delete("cas")

stephenrlouie: just need to make sure we never commit the "cas" value is what I meant. Ensure all inserts or updates will Delete

manager/manager.go
adamschaub: Yes, it will always contain one entry for this service's plugin name. 

It wouldn't need two entries. `[]int{1}[1:] == []int{}`. So one entry means the list would be empty.

These are being removed right now because any generated vars will cause the vars to be detected as changed, causing the plugin to run an update. I need to update this to more specifically remove just the generated vars of the parent service, so we can still trigger updates when statically assigned vars are changed in the service.

manager/manager.go
adamschaub: No, json marshal orders the keys in ascending order: https://play.golang.org/p/13EBUCnqoo

database/nulldb.go
adamschaub: We could, but I've kept it separate here because not all CAS is created equal. Couchbase internally tracks it's own cas values, nulldb stores it with the object. I figured it makes more sense to namespace it under nulldb.

manager/manager.go
adamschaub: Name is unique enough because vars are defined per-plugin

Makefile
```diff
@@ -53,3 +53,8 @@ fast:
 container:
 	docker run -t -w /go/src/github.com/sigma-dev/sigma -v `pwd`:/go/src/github.com/sigma-dev/sigma golang:1.7.1-wheezy make
 	docker build -t $(SIGMA_REPO):$(SIGMA_TAG) .
+
```
stephenrlouie: do not need

api/v1/convert/convert.go
```diff
@@ -12,62 +12,44 @@ import (
 
 func FromModelServiceOperation(o model.ServiceOperation) apiclient.ServiceOperation {
 	// parse all sites
-	sites := []apiclient.Site{}
-	for _, s := range o.Sites {
-		sites = append(sites, FromModelSite(s))
-	}
+	// TODO: How to populate sites using ModelRefs
```
stephenrlouie: lets make an issue so we remember to follow up

Makefile
mikeynap: if this gets deleted, i will rebase master to be just this make command.

api/v1/convert/convert.go
stephenrlouie: https://github.com/sigma-dev/sigma/issues/285

api/v1/convert/convert.go
stephenrlouie: Nah. I mean the model structs having a Name field on top of the MetaName. Couldn't we just have the MetaImpl, I would understand why the [API](https://github.com/sigma-dev/sigma/blob/rebirth/api/v1/client/plugin.go#L15) would have it

- [Plugin](https://github.com/sigma-dev/sigma/blob/rebirth/model/model.go#L41)
- [Service](https://github.com/sigma-dev/sigma/blob/rebirth/model/model.go#L81)
- and so on....

api/v1/convert/convert.go
adamschaub: Being addressed by @vgunning in #269

plugin/client/plugin_test.go
```diff
@@ -3,119 +3,192 @@
...
-		t.Error(err)
-	}
-	b, e := ioutil.ReadFile(path.Join(SaveDir, "output"))
-	if e != nil {
-		t.Error(e)
-	}
-	if !bytes.Contains(b, []byte("victoria")) {
-		t.Error("File not written.")
-	}
+// TODO: Removed for now because output adding has been simplified
```
stephenrlouie: Delete this then?

plugin/client/plugin.go
```diff
@@ -58,128 +70,142 @@ func GetArg() PluginArg {
...
+		return err
+	}
+	_, err = f.WriteString(p)
+	return err
 }
 
-func getInputs(src string) (map[string]interface{}, error) {
-	b, err := ioutil.ReadFile(path.Join(inputSavedir, src))
+// WriteVars takes Vars and marshals/writes them to the output file
+// TODO - collapse "input" and "output" into one file? (keeping separate for now for debugging purposes)
```
stephenrlouie: Issue made: https://github.com/sigma-dev/sigma/issues/293

task/nomad.go
```diff
@@ -266,16 +262,22 @@ func (nt *NomadTasker) chooseNomadNode(sites []model.Site) (string, error) {
...
-		artifacts[i] = &nomadapi.TaskArtifact{
-			GetterSource: stringToPtr(m[k]),
-			RelativeDest: &secretDir,
+	artifacts := make([]*nomadapi.TaskArtifact, 0, len(svcVars))
+	for _, list := range svcVars {
+		for _, vars := range list {
+			for _, v := range vars {
+				if v.Secret {
+					artifacts = append(artifacts, &nomadapi.TaskArtifact{
+						// TODO: Might be risky? Should always pass through toPluginVars though
```
stephenrlouie: what?

task/nomad_test.go
```diff
@@ -33,21 +33,22 @@ func TestNomadCreateAndRun(t *testing.T) {
...
 
 	// test create
-	testOp := sampleServiceOperation()
-	m.EXPECT().NewServiceJob(testOp.Id, testOp.Id, "global", 1).Return(&nomadapi.Job{})
-	m.EXPECT().NewTaskGroup(testOp.Service.Name, 1).Return(&nomadapi.TaskGroup{})
-	testJob, err := nt.CreateNomadJob(testOp)
+	testOp, sites := sampleServiceOperation()
+	m.EXPECT().NewServiceJob(testOp.Id, testOp.GetID(), "global", 1).Return(&nomadapi.Job{})
+	m.EXPECT().NewTaskGroup(testOp.Service.GetName(), 1).Return(&nomadapi.TaskGroup{})
+	// TODO: Come back, add sites?
```
stephenrlouie: https://github.com/sigma-dev/sigma/issues/294

api/v1/server/server.go
```diff
@@ -925,7 +871,8 @@ func (s *Server) GetServiceOperations(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	//query all operations for this service
-	modOps, getErr := s.db.QueryServiceOperations(model.ServiceOperation{Service: svc}, nil)
+	// TODO: This query doesn't work. Need to either tag old operations with the serviceID or find a way to query by modelRef
```
stephenrlouie: https://github.com/sigma-dev/sigma/issues/289

database/nulldb.go
```diff
@@ -152,22 +148,40 @@ func query(mdaters []model.Metadataer, m model.Metadataer, filters map[string]in
 			} else {
 				parts := strings.Split(k, ".")
 				field := rv
+				// TODO: Matching a MetaDater won't work because .Name in json --> MetaName.
```
stephenrlouie: https://github.com/sigma-dev/sigma/issues/289

manager/manager.go
```diff
@@ -242,25 +239,26 @@ func (m *manager) cleanSites() error {
...
 			continue
 		}
 		if (len(childrenName) + len(childrenID)) > 0 {
 			continue
 		}
 
-		//check if I have services
-		svcsName, err := m.db.QueryServices(model.Service{SiteName: myself.Name}, nil)
+		// Query does not work with modelRef... use query map instead.
+		// TODO: look into alternative methods.
```
stephenrlouie: https://github.com/sigma-dev/sigma/issues/289

manager/manager.go
```diff
@@ -570,13 +726,14 @@ func (m *manager) finishOperation(op model.ServiceOperation) (retErr error) {
...
 	// Stop the active serviceOperation
 	if err := m.tasker.Remove(&op); err != nil && err != task.ErrServiceNotFound {
-		log.WithError(err).Errorf("Unable to tasker.Remove %s", op.Id)
+		log.WithError(err).Errorf("Unable to tasker.Remove %s", op.GetID())
 	}
 
 	// get all serviceOperations for the service corresponding to this serviceOperation
-	existingOps, err := m.db.QueryServiceOperations(model.ServiceOperation{Service: model.Service{Id: op.Service.Id}}, nil)
+	// TODO: This query is busted
+	existingOps, err := m.db.QueryServiceOperations(model.ServiceOperation{Service: model.ModelRef{}.WithID(op.Service.GetID())}, nil)
```
stephenrlouie: https://github.com/sigma-dev/sigma/issues/289

manager/manager.go
```diff
@@ -585,349 +742,128 @@ func (m *manager) archiveOperation(op model.ServiceOperation) {
...
-		} else if item.Value != nil {
-			if conv, ok := model.ConvertIO(item.Value, output.Type, plugin); ok {
-				output.Actual = conv
-				output.Set = true
-				changed = true
-			} else {
-				log.Warnf("Service %s output %s cannot use value %#v, accepts type %s", svc.Name, output.Name, output.Actual, output.Type)
+	for name, v := range output {
+		if svcVar, ok := svc.Vars[name]; ok {
+			// TODO: Only set if svcVar.Generated == true?
```
stephenrlouie: https://github.com/sigma-dev/sigma/issues/290

model/metadata.go
```diff
@@ -4,22 +4,74 @@ package model
 import (
 	"fmt"
 	"strconv"
+	"time"
+)
+
+// Common metadata keys used by sigma
+const (
+	// TODO: Get databases to use these tags when searching for Id/Name
```
stephenrlouie: https://github.com/sigma-dev/sigma/issues/289

plugin/client/plugin.go
```diff
@@ -22,9 +22,22 @@ var (
...
+		Value      interface{} `json:"value,omitempty"`
+		Secret     bool        `json:"secret,omitempty"`
+		SecretPath string      `json:"secret_path,omitempty"`
+	}
+	ServiceVars map[string][]map[string]ServiceVar
+	Vars        map[string][]map[string]interface{}
 )
 
 // parse savedir environment variables + init defaults if unset
+// TODO - condense input and output savedir into one savedir?
```
stephenrlouie: https://github.com/sigma-dev/sigma/issues/293

PR #269, opened 2017-07-20
[WIP] Internal ID object consistency

### LINKED ISSUE #208: DB object reference consistency

### LINKED ISSUE #285: Populate Sites using Model Refs in FromModelServiceOperation
Tracking a code TODO:

`convert.go` To properly take a model.ServiceOperation -> API.ServiceOperation and incorporate Sites. today it is commented out.

### LINKED ISSUE #287: Clean up model.Name duplication
##### Description

Many objects in the model have a name field. Such as `model.Service.Name`, it also has a `model.Service.MetaImpl["Name"]` we should remove the duplicate `.Name` field and only use the `MetaImpl["Name"]`. 

Anything with a MetaImpl should not have a Name or ID. 

Being fixed by: #269

### LINKED ISSUE #289: Query With Model Ref
##### Description

TODO from Line 255 in manager.go, QueryingServices.`Query does not work with modelRef... use query map instead`

vgunning:
### PR checklist

### Change log

- [x] Change all internal references to IDs instead of Names
- [x] Convert ID references to name references when returning API objects
- [x] Change internal site parent site references to the object ID
- [ ] Make sure that ID is immutable (still mutable because of how service operations are handled, but is its own type)
- [x] Make sure that plugins with same name and different versions are allowed

### Notes

Resolves #208 
Resolves #289 
Resolves #287 
Resolves #285


model/refs.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (mt MetaType) String() string {
+	return string(mt)
+}
+
+func ToMetaType(t string) MetaType {
+	return MetaType(t)
+}
+
+type MetaID string
```
adamschaub: why is this it's own type? Same with MetaType. Might as well just make them strings and avoid the potential casting issues.

manager/manager.go
```diff
@@ -982,15 +982,13 @@ func (m *manager) gatherChildSitesWorker(originID, parentID string, isFirst bool
...
+// Get plugin from service ref
 func (m *manager) pluginFor(svc model.Service) (model.Plugin, error) {
-	plugin, err := m.db.GetPlugin(svc.PluginRef.GetID().String())
-	svc.PluginRef.SetID(plugin.GetID())
-	return plugin, err
+	return m.db.GetPlugin(svc.PluginRef.GetID().String())
 }
 
 func operationID(svcID string) string {
-	return fmt.Sprintf("%s-operation", svcID)
```
adamschaub: why change?

PR #270, opened 2017-07-21
New Functional Tests

mikeynap:
See the functional test case of https://github.com/sigma-dev/sigma/blob/master/tests/plan.md
- Functional Tests that run in Dev/Docker/Nomad[lol] mode.
- Supports starting/stopping sigma directly if `go test GitHub.com/sigma-dev/sigma/tests/functional` is used. 
- The `go test -run` command can be used to specify which test cases to run. `go test -run TestDev` runs all the subtests in Dev configuration. A specific subtest can be run with `go test -run TestDev/SubTestName`. Regexes can be used as normal.
- Running `go test GitHub.com/sigma-dev/sigma/tests/functional/sigtest` assumes sigma has been started externally. It is up to the caller to start/stop the sigma instances between runs. Also expects an -address flag to specify the endpoint to hit. Since this mode/package does not directly start/stop/clear sigma/thedb, the -run argument is required (otherwise it would try to run all three modes back to back, but that obviously would fail since sigma would not have been shutdown between calls). 

The tests also support a few cases for running with an external orchestrator:
- An optional -wait flag can be provided to wait before the major tests cases (dev/docker,etc) so that some other service can restart/reconfigure the sigma instance between calls. If doing some automated deployment, setting wait to be a few minutes should be sufficient for the deployer to reconfigure sigma. 
- An optional -notify flag can be provided to hit an endpoint when one of the major test cases is about to start. This can be used to trigger a specific configuration of sigma. The url will include a test=testThatWillRun query parameter.
- If these are both used, the -run parameter should not be required (one should be able to run all three tests with an external sigma since it will be orchestrated.).
- An alternate strategy is to not supply a -wait parameter and have the endpoint keep the initial status post alive until it has setup sigma.


cmd/sigma.go
```diff
@@ -109,18 +109,19 @@ func sigmaRun(cmd *cobra.Command, args []string) {
...
 		}
 	}
-
+	taskrToken := auth.TaskerToken(tokener).String()
 	// Tasker
 	var tskr task.Tasker
 	switch gCfg.Tasker {
 	case "nomad":
-		if tskr, globalErr = task.NewNomadTasker(gCfg.Nomad, auth.TaskerToken(tokener).String()); globalErr != nil {
+
```
stephenrlouie: NIT

manager/manager.go
```diff
@@ -245,39 +244,32 @@ func (m *manager) cleanSites() error {
 		//check if I have children
 		childrenName, err := m.db.QuerySites(model.Site{Parent: myself.Name}, nil)
 		if err != nil {
-			log.Warnf("GC failed to get childrenSites for: %q. %v", myself.Name, err)
```
stephenrlouie: I could get toning down the logs, but I think we should print something if we are getting errors. Not a fan of silence given we know there is a DB issue.

manager/manager.go
```diff
@@ -371,7 +363,7 @@ func (m *manager) updateService(serviceID string) {
 	if ok, changed, old := m.inputSatisfied(&service, plugin); !ok {
 		log.Debugf("Inputs for service %s not yet satistifed- service not ready", service.Name)
 		service.StatusDescription = "Waiting for inputs to be ready"
-		if err := m.db.UpdateService(service); err != nil {
+		if err := m.db.UpdateService(service); err != nil && err != database.ErrCas {
```
stephenrlouie: should we retry the update here if its an `ErrCas` we also have many other instances of `db.UpdateService(` where we do not make the `err` distinction.

tests/functional/deploy.sh
```diff
@@ -1,20 +1,11 @@
...
-sed -i "s/localhost:5672/${MQIP}:5672/g" ${SIGMAPATH}/tests/functional/config.yaml
-sed -i "s/10s/1s/g" ${SIGMAPATH}/tests/functional/config.yaml
-sed -i "s/2m/1s/g" ${SIGMAPATH}/tests/functional/config.yaml
-docker stop sigma-functional
-docker rm sigma-functional
-docker run -d --name sigma-functional -p 24601:24601 -v /var/lib/sigma:/var/lib/sigma -v ${SIGMAPATH}/tests/functional:/sigma -v /var/run/docker.sock:/var/run/docker.sock sigmadev/sigma:latest --config /sigma/config.yaml
-echo "sigma functional container"
+${SIGMAPATH}/tests/integration/auth/ldap.sh stop
+${SIGMAPATH}/tests/integration/auth/ldap.sh
+
```
stephenrlouie: NIT

tests/functional/deploy.sh
```diff
@@ -1,20 +1,11 @@
...
-sed -i "s/127.0.0.1:24601/0.0.0.0:24601/g" ${SIGMAPATH}/tests/functional/config.yaml
-MQIP=$(docker inspect --format '{{ .NetworkSettings.IPAddress }}' sigma-queue)
-sed -i "s/localhost:5672/${MQIP}:5672/g" ${SIGMAPATH}/tests/functional/config.yaml
-sed -i "s/10s/1s/g" ${SIGMAPATH}/tests/functional/config.yaml
-sed -i "s/2m/1s/g" ${SIGMAPATH}/tests/functional/config.yaml
-docker stop sigma-functional
-docker rm sigma-functional
-docker run -d --name sigma-functional -p 24601:24601 -v /var/lib/sigma:/var/lib/sigma -v ${SIGMAPATH}/tests/functional:/sigma -v /var/run/docker.sock:/var/run/docker.sock sigmadev/sigma:latest --config /sigma/config.yaml
-echo "sigma functional container"
+${SIGMAPATH}/tests/integration/auth/ldap.sh stop
```
stephenrlouie: should these ldap scrips and nomad deploy live somewhere else that isn't under `test/integration` since we are using them here too.

tests/functional/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,444 @@
...
+		child := child
+		if !strings.Contains(child, "tcp") && !strings.Contains(child, ":") {
+			child = fmt.Sprintf("tcp://%v:%d", child, c.DockerPort)
+		} else if !strings.Contains(child, "tcp") {
+			child = fmt.Sprintf("tcp://%v", child)
+		} else if strings.Contains(child, ":") {
+			child = fmt.Sprintf("%v:%d", child, c.DockerPort)
+		}
+
+		cl, err := client.NewClient(child, "1.24", nil, nil)
```
stephenrlouie: can we get a configurable version for docker?

tests/functional/sigtest/config.go
```diff
@@ -0,0 +1,92 @@
...
+func (t TestConfig) APIClientUser() (*apiclient.DefaultApi, error) {
+	cl := t.APIClient()
+	tok, _, err := cl.Login("testing", "123CISCOBOSCO123")
+	if tok == nil {
+		return nil, err
+	}
+	return cl, err
+}
+
+func (t TestConfig) APIClientAdmin() (*apiclient.DefaultApi, error) {
```
stephenrlouie: whats the difference b/t this and the above?

tests/functional/sigtest/login.go
```diff
@@ -0,0 +1,99 @@
...
+	"fmt"
+	"net/http"
+	"testing"
+
+	apiclient "github.com/sigma-dev/sigma/api/v1/client"
+	"github.com/sigma-dev/sigma/auth/user"
+)
+
+/*
+	These tests assume the user admin/cisco123 & testing/123CISCOBOSCO123  already exists
```
stephenrlouie: can we make these consts someplace that everyone can ref.

tests/functional/sigtest/plugins_services.go
```diff
@@ -0,0 +1,298 @@
...
+	_, file, _, _ := runtime.Caller(1)
+	file = path.Join(path.Dir(file), "json", filename)
+	b, err := ioutil.ReadFile(file)
+	if err != nil {
+		return err
+	}
+	return json.Unmarshal(b, i)
+}
+
+func testPluginServicesCRUD(t *testing.T) {
```
stephenrlouie: Could we break everything down into its CRUD. This being C and the following being R.

tests/functional/sigtest/run.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	RunLoginTests(t, cfg)
+	RunSitesTests(t)
+	RunPluginServicesTests(t)
+
+	if standalone {
+		sigint()
+		select {
+		case <-channel:
```
stephenrlouie: gofmt?

tests/functional/sigtest/run.go
```diff
@@ -0,0 +1,232 @@
...
+		"--couchbase-mb-quota", "512", "--couchbase-timeout", "32s")
+	return cbClear
+}
+
+func setupAuthLDAP(cfg *TestConfig) {
+	cfg.AuthMode = "ldap"
+	cfg.Args = append(cfg.Args, "--auth-method", "ldap", "--auth-ldap-admin-distinguished-name", "cn=admin,dc=sigma-dev,dc=github,dc=io", "--auth-ldap-admin-password", "cisco123", "--auth-ldap-host", "127.0.0.1")
+}
+
+func setupAuthDB(cfg *TestConfig) {
```
stephenrlouie: should these functions also start their containers?

tests/functional/deploy.sh
```diff
@@ -1,20 +1,11 @@
 echo "Functional test setup"
```
stephenrlouie: should we provide some intelligence here so you could locally replicate the set ups Dev/Docker/Nomad? We can leverage these for the remote testing too if we wanted.

tests/functional/README.md
```diff
@@ -0,0 +1,37 @@
+# Sigma Functional Tests
+
+## Test Methods
+The functional tests can be run in three different configurations:
+
+### Local Testing
+`go test github.com/sigma-dev/sigma/tests/functional `
+This method starts and runs sigma locally, automatically restarting/resetting between runs.
+The runner should ensure the necessary services are started (use `tests/functional/deploy.sh`).
```
stephenrlouie: they should just get a local sigma running in the configuration they desire to test

tests/functional/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+### TestDev
+`-run TestDev`
+Run the tests with a "dev" configuration (nulldb, channel-init, auth:none, docker tasker)
+
+### TestDocker
+`-run TestDocker`
+Run the tests with a "docker" configuration (couchbase, rabbitmq, auth:db, docker tasker)
+
+### TestNomad
```
stephenrlouie: lets get a nomad config to make running local easier

tests/functional/sigtest/run.go
```diff
@@ -0,0 +1,232 @@
...
+		"--auth-shared-secret", "sigmafunc")
+}
+
+func setupHTTPS(cfg *TestConfig) {
+
+}
+
+func setupNomad(cfg *TestConfig) {
+	cfg.Nomad = true
+	cfg.Args = append(cfg.Args, "--tasker", "nomad", "--nomad-client-url", cfg.apiURL(4646), "--nomad-api-url", cfg.APIURL)
```
stephenrlouie: can we do something more intelligent. for local we need it to be the nomad VM IP.

tests/functional/sigtest/run.go
```diff
@@ -0,0 +1,238 @@
...
+	time.Sleep(tm)
+}
+
+func RunTests(t *testing.T, cfg TestConfig, standalone bool) {
+	channel := make(chan bool)
+	if standalone {
+		c := cmd.SigmaCmd
+		c.SetArgs(cfg.Args)
+		sigmaconfig.InitFlags(c)
+		go func() {
```
stephenrlouie: is there a good way to clean up mid-test?

tests/functional/stop.sh
```diff
@@ -1,8 +1,6 @@
 echo "Functional test cleanup"
 SIGMAPATH=${GOPATH}/src/github.com/sigma-dev/sigma
 ${SIGMAPATH}/database/stop.sh
+${SIGMAPATH}/tests/integration/auth/ldap.sh stop
```
stephenrlouie: ./tests/functional/stop.sh: line 4: /Users/stelouie/Documents/Go/src/github.com/sigma-dev/sigma/tests/integration/auth/ldap.sh: No such file or directory
./tests/functional/stop.sh: line 5: /Users/stelouie/Documents/Go/src/github.com/sigma-dev/sigma/tests/integration/nomad_stop.sh: No such file or directory

tests/functional/sigtest/run.go
```diff
@@ -0,0 +1,243 @@
...
+			cmd.SigmaCmd.Execute()
```
stephenrlouie: Err catch?

tests/functional/sigtest/run.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+	time.Sleep(tm)
+}
+
+func RunTests(t *testing.T, cfg TestConfig, standalone bool) {
+	channel := make(chan bool)
+	if standalone {
+		c := cmd.SigmaCmd
+		c.SetArgs(cfg.Args)
+		sigmaconfig.InitFlags(c)
```
stephenrlouie: err catch?

PR #275, opened 2017-07-25
Fixed some auth bugs/potential bugs

mikeynap:
- Fixes a bug with tokenizer (the order of initialization was messed up in a merge).
- Fixes a potential crash in Couchbase
- Fixes a potential bug in QuerySecrets


auth/authorizer/db.go
```diff
@@ -89,7 +89,7 @@ func (a *DBAuth) AddUser(un, pw string, role user.Role) error {
 	}
 
 	for _, l := range un {
-		if unicode.IsSymbol(l) || unicode.IsSpace(l) {
+		if !unicode.IsLetter(l) {
```
adamschaub: won't allow numbers in usernames? Or special symbols (thinking about `@` for emails-as-usernames)

database/couchbase.go
```diff
@@ -649,9 +649,11 @@ func addQuery(params map[string]interface{}, query string, isFirst bool, f refle
 		f = f.Elem()
 	}
 	if (f.Kind() < reflect.Array || f.Kind() == reflect.String || f.Kind() == reflect.Interface) && len(fmt.Sprintf("%v", f)) != 0 {
-		key := randomID()
-		query, isFirst = queryHelper(isFirst, query, fieldName+"=$"+key)
-		params[key] = f.Interface()
+		if f.CanInterface() {
```
adamschaub: good catch

database/couchbase.go
```diff
@@ -790,6 +797,7 @@ func (cb *Couchbase) QuerySecrets(ss model.Secret, filters map[string]interface{
 	var listRet []string
 	for rows.Next(&retN) {
 		for _, n := range retN {
+			n := n
```
adamschaub: close one

database/couchbase.go
```diff
@@ -780,7 +782,12 @@ func (cb *Couchbase) UpdateSecret(s model.Secret) error {
 }
 
 func (cb *Couchbase) QuerySecrets(ss model.Secret, filters map[string]interface{}) ([]string, error) {
-	ss.Set("Type", ss.Type())
+	if filters == nil {
```
adamschaub: Why do we need these steps, doesn't makeQuery add metadata filters automatically? The filter get overwritten by the presence of metadata in the object anyway

mikeynap: Secrets are ONLY queried by the filter map. (you can see we pass nil to makeAndExecuteQuery). 

Thus the map needs to contain the Type.

adamschaub: ah, didn't even see that. Good stuff

PR #295, opened 2017-08-02
Shipping Overhaul

adamschaub:
### PR checklist

- [x] Plugin Client
- [x] Plugins
- [x] API Definitions
- [x] Manager
- [x] Tasker
- [x] FIX TESTS

### Notes

Replace mounted files as mechanism for interacting between manager/tasker & service operations. All communication will be API driven. Manager/Tasker provides env vars to be used inside plugin to make appropriate callbacks.


api/v1/server/routers.go
```diff
@@ -24,96 +24,28 @@ type Route struct {
...
 
 func newRouter() *Router {
-	return &Router{gmux: mux.NewRouter().StrictSlash(false)}
-
+	return &Router{mux.NewRouter().StrictSlash(false)}
 }
 
 type Routes []Route
 
-func (r *Router) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
```
mikeynap: ummmmmmmm why did you delete this? Please restore, this is for cross origin?

api/v1/server/routers.go
```diff
@@ -24,96 +24,28 @@ type Route struct {
 }
 
 func newRouter() *Router {
-	return &Router{gmux: mux.NewRouter().StrictSlash(false)}
```
mikeynap: restore.

api/v1/server/routers.go
```diff
@@ -125,106 +57,122 @@ func (r *Router) InitRoutes(s *Server) {
...
 		Route{
-			Name:       "GetServiceOperationOutputs",
+			Name:       "GetServiceOperationPreviousInput",
 			Method:     "GET",
-			Pattern:    "/v1/services/{service-name}/operation/{service-operation-id}/outputs",
-			Handler:    s.GetServiceOperationOutputs,
-			Middleware: []Middleware{s.Auth},
+			Pattern:    "/v1/services/{service-name}/operation/{service-operation-id}/previous-input",
+			Handler:    s.GetServiceOperationPreviousInput,
+			Middleware: []Middleware{},
```
mikeynap: why are all of the s.Auths gone?

api/v1/server/routers.go
```diff
@@ -290,15 +280,15 @@ func (r *Router) InitRoutes(s *Server) {
...
+	entry := r.
 		Methods(route.Method).
 		Path(route.Pattern).
 		Name(route.Name)
 
 	if route.Handler != nil {
 		handler := route.Handler
-		for i := len(route.Middleware) - 1; i >= 0; i-- {
-			handler = route.Middleware[i](handler).(http.HandlerFunc)
+		for _, mw := range route.Middleware {
```
mikeynap: ???????????

config.yaml
```diff
@@ -6,7 +6,7 @@ Manager:
   ServiceStatusInterval: 20s
 Docker:
   SaveDir: /var/lib/sigma
-  APIUrl: 127.0.0.1:24601
+  APIUrl: 10.131.32.49:24601
```
mikeynap: make sure this is reverted b4 merge.

config.yaml
```diff
@@ -22,5 +22,5 @@ Rabbit:
   Password: guest
   URL: localhost:5672
 Server:
-  Host: 127.0.0.1:24601
+  Host: 10.131.32.49:24601
```
mikeynap: same

plugin/client/plugin.go
```diff
@@ -109,8 +105,57 @@ func (v Vars) UnmarshalVars(plugin string, idx int, i interface{}) error {
...
+		for _ = range ticker.C {
+			p.Heartbeat()
+		}
+	}()
+}
+
+// Send a heartbeat for this plugin - allows Sigma to increase the TTL for this plugin
+// Every plugin decides how often to Heartbeat
+func (p *PluginClient) Heartbeat() error {
+	return p.sigmaAPIPost(nil, fmt.Sprintf("%s/%s", p.sigmaURL, "heartbeat"))
```
mikeynap: what happens if this is NOT called? will things still work?>

task/nomad.go
```diff
@@ -195,44 +183,27 @@ func (nt *NomadTasker) createPluginTask(so *model.ServiceOperation) (*nomadapi.T
...
-	it.DestPath = stringToPtr("local/input")
-	it.EmbeddedTmpl = stringToPtr(string(jInput))
-
-	pit := nomadapi.Template{}
-	prevVars := toPluginVars(so.PreviousVars, env[SigmaEnvSecretDir], "previous-input-")
-	pjInput, _ := json.Marshal(prevVars)
-	pit.DestPath = stringToPtr("local/previous-input")
-	pit.EmbeddedTmpl = stringToPtr(string(pjInput))
-	templates := []*nomadapi.Template{&it, &pit}
+	env[SigmaEnvAuthToken] = "" // TODO: Get token
```
mikeynap: <----

api/v1/server/routers.go
mikeynap: plz  revert

mikeynap:
I would personally for the test just use this (bind to this address, pass this address to the docker tasker).
```
// Get preferred outbound ip of this machine
func GetOutboundIP() net.IP {
	conn, err := net.Dial("udp", "8.8.8.8:80")
	if err != nil {
		log.Fatal(err)
	}
	defer conn.Close()

	localAddr := conn.LocalAddr().(*net.UDPAddr)
	return localAddr.IP
}
```

api/v1/server/routers.go
adamschaub: no

config.yaml
adamschaub: cha cha

config.yaml
adamschaub: yes

plugin/client/plugin.go
adamschaub: Heartbeat is a large (relatively) time frame. It exists only to catch tasks that have somehow deadlocked or failed without warning. Our default now is like 5m. If your plugin takes longer than that to complete, and it doesn't call heartbeat, then yes it will fail.

task/nomad.go
adamschaub: ^?

api/v1/server/routers.go
adamschaub: Because I borked it.

task/nomad.go
mikeynap: We need to get the token in here before merge.

PR #305, opened 2017-08-07
Run Nomad Jobs on datacenters, not nodes 

meganokeefe:
### changelog 

- [x] a serviceOperation now has a single `model.Site` instead of a slice 
- [x] if a Site has children, manager chooses one at random (including the parent in the pool) to deploy on 
- [x] tasker.Create now takes in a single Site instead of a slice 
- [x] in `createNomadJob`, removed the node constraint. No longer defaulting the datacenter to dc-local, but instead going into site attributes and grabbing `datacenter` string. This means that all services must provide a site name-- 
- [x] -- because there is no longer any site `local`. 
- [x] fixed up the nomad integration tests -- since we no longer have `tasker.Status`, now they just test `Create`, `isRunning`, and `Remove.`
- [x] added a multi-datacenter vagrant setup, `multi-dc`. Sets up 2 datacenters, `dc-1` and `dc2`, each with 2 clients. All fall under the same server, `server-1`. This lets you have multiple sites in sigma, each pointing to different datacenters, so that you can run services on different sites. 

### how to test 

1. stand up multiple client and server VMs with `./vagrant/nomad/multi-node`. You should be able to do a `nomad node-status` from the server and see `x` client nodes, where `x` is NUM_CLIENTS. Ssh into the clients and watch the logs-- this is how you can see which client gets the allocation to run the plugin container. Unfortunately just doing a `nomad alloc-status` won't tell you which client the Server sent it to. 
2. add a site where `Attributes["datacenter"] = "dc-local"` (see ./examples/helloWorld/addSite.sh) 
3. stand up sigma, add a plugin 
4. add a service and go into your client vms to watch the magic happen 

(If you want to be crazy you can also use `multi-dc`and have sites with `dc-1` and `dc-2` respectively.) 

### todo 

add customization for `multi-dc`


tests/integration/task/nomad/nomad_test.go
```diff
@@ -183,71 +186,57 @@ func TestNomadCreate(t *testing.T) {
...
+	err := nt.Create(op, model.Site{Name: "sanfrancisco", Attributes: attr})
+	if err != nil {
+		t.Error(err)
+	}
 
-// calls tasker.Status <maxRetry> times, looking for status = expStatus
-func statusWithRetry(op model.ServiceOperation, expStatus bool) error {
-	for i := 0; i < maxRetry; i++ {
-		stat, err := nt.IsRunning(op)
+	time.Sleep(time.Second * 3)
```
stephenrlouie: could be hardware dependent? can we just poll nomad a bunch and watch it transition into the running state?

tests/integration/task/nomad/nomad_test.go
```diff
@@ -183,71 +186,57 @@ func TestNomadCreate(t *testing.T) {
...
-		if err != nil && !strings.Contains(err.Error(), "allocations") { //ignore a "no allocations" err
-			return err
-		}
-		if stat == expStatus {
-			fmt.Printf("\n------> Got expected tasker running (%s) after %d retries", stat, i+1)
-			return nil
-		}
-		fmt.Printf("\nFor Op=%s Got status=%s And Events: %#v, \n", op.Id, stat, ev)
-		time.Sleep(time.Millisecond * 200)
+	time.Sleep(time.Second * 3)
```
stephenrlouie: same as above, lets poll.

examples/hello-world/add-site.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+curl -H "Content-Type: application/json" --data @centralsite.json http://172.16.7.1:24601/v1/sites
```
stephenrlouie: s/172..../127.0.0.1

stephenrlouie: s/172.../127.0.0.1

meganokeefe:
reminder to remove the docker savedir and nomad API address from the Config.

manager/manager.go
```diff
@@ -936,12 +912,15 @@ func (m *manager) secretURLFromToken(t string) string {
 
 // recursive helper for newOperation, goes down the site tree to get all child Sites plus myself
 //the reason for the helper is so we know if we are calling gatherChildSitesWorker for the first time (with the root)
-func (m *manager) gatherChildSites(parentSite model.ModelRef) ([]model.ModelRef, error) {
+func (m *manager) chooseChildSite(parentSite model.ModelRef) (model.ModelRef, error) {
```
stephenrlouie: we might have to make this part of a tasker interface, because we need child sites that all have `Attributes[datacenter]` and whatever `NomadTasker` needs to run. Right now this will not run with a parent site, I set up `site1` -> `global`. Passing in `global` as the site for a Service does not work. 

Logs:
```
ERRO[2017-08-09T14:41:51-04:00] Could not start operation c1c4fe52-d428-48ae-91aa-eb97ecb3869b-operation  error="Site provided has no datacenter Attribute" file=manager.go line=633
```

stephenrlouie: The reason for interfacing it, would be other taskers might want children with other different attributes. So one true child finder might not work.

task/nomad.go
```diff
@@ -128,19 +119,14 @@ func (nt *NomadTasker) Shutdown() {
...
-	// choose a random nomad node ID from sites. if site = Local, provide no constraints for dc-local
-	nomadNodeID, err := nt.chooseNomadNode(sites)
-	if err != nil {
-		return nil, err
-	}
-	if nomadNodeID != "" {
-		constraint := nt.Wrapper.NewConstraint("${node.unique.id}", "=", nomadNodeID)
-		j.Constraints = []*nomadapi.Constraint{constraint}
+	// List all sites, as long as they have the datacenter field. TODO - change this behavior?
+	if _, ok := s.Attributes["datacenter"]; !ok {
```
stephenrlouie: this is where the parent issue gets kicked off, because we end up with a site without the `datacenter` attribute here.

stephenrlouie:
if the site is not present before the service is added, I end up looping on this. https://github.com/sigma-dev/sigma/blob/datacenters/manager/manager.go#L788
```
ERRO[2017-08-09T16:18:05-04:00] Could not operate service helloSvc            error="Object not found" file=manager.go line=788
ERRO[2017-08-09T16:18:07-04:00] Could not operate service helloSvc            error="Object not found" file=manager.go line=788
ERRO[2017-08-09T16:18:09-04:00] Could not operate service helloSvc            error="Object not found" file=manager.go line=788
ERRO[2017-08-09T16:18:11-04:00] Could not operate service helloSvc            error="Object not found" file=manager.go line=788
```

stephenrlouie:
^ fixed.... https://github.com/sigma-dev/sigma/blob/datacenters/api/v1/server/validate.go#L165-L166

PR #309, opened 2017-08-10
Service Attributes, ChooseSite, HA Nomad

### LINKED ISSUE #307: Tasker GatherSiteChildren Issue for Parent sites without required attributes
##### Description

Need to update how to [GatherChildrenSites](https://github.com/sigma-dev/sigma/blob/master/manager/manager.go#L869) Found during review for https://github.com/sigma-dev/sigma/pull/305 and decided it might be a decent amount of work. If a parent site is created that does not contain `datacenter`, say `EAST` is a parent to `BOS` and `NYC`. `BOS` & `NYC` both have the datacenter attribute, while `EAST` is simply creating a hierarchy. With the current Gather function we can end up with EAST, it will get rejected by the nomad tasker.

##### Sigma Configuration
nomad.yaml found in the PR mentioned

##### Steps to reproduce

1. checkout the branch https://github.com/sigma-dev/sigma/pull/305
2. start the nomad vms `cd vagrant/nomad/multi-dc`
3. `vagrant up`
4. run sigma with `./bin/sigma --config nomad.yaml`

##### Suggest a Fix

We make gatherChildSites part of the interface for each tasker to implement. so it can look at a site, follow the potential tree and find what is a "runnable" site for itself. Maybe nomad tasker requires `datacenter` but another tasker may want another field in `attributes` so we can no longer have a generic site finder for all taskers since it depends on `attributes` related to that tasker implementation.

meganokeefe:
Closes #307  

### changelog 
- [x] added an `Attributes` field (`map[string]interface`) to `Service` and `ServiceOperation`. Represents additional metadata/constraints for a given Service, to be used (or ignored) by the tasker. 
- [x] nomadTasker handles Attributes in `Create()` by gathering all Site Attributes then combining (and overriding) them with the ServiceOperation Attributes. Then creates Job Constraints based on Nomad's allowed Interpolated Variables: https://www.nomadproject.io/docs/runtime/interpolation.html#interpreted_node_vars   

Example nomad-readable Attribute (if no operand is listed, defaults to "=") 
```
Attributes: {
    "${attr.kernel.version}": { 
          "value": "1.0", 
          "operand": ">=", 
    }
}
```

- [x] added `ChooseSite` to the tasker interface, which will be called from Manager. `tasker.Create()` now takes in one Site, which is the result of ChooseSite. 
- [x] `nomadTasker.ChooseSite` parses out the sites with a `datacenter` attribute and chooses a site at random. DockerTasker chooses a site at random without constraints.

- [x] multinode vagrant nomad environment is now properly HA Server, by adding the `server-join` command and giving clients a list of server IPs to register with. 
- [x] unit tests 
- [x] integration tests 

### testing Nomad HA Servers 

vagrant up in `multinode` and have `NUM_SERVERS` > 1. 
when it's done (this example has 2 servers), should be able to do this: 
```
[vagrant@server-1 ~]$ nomad server-members
Name             Address       Port  Status  Leader  Protocol  Build  Datacenter  Region
server-1.global  172.16.7.101  4648  alive   true    2         0.6.0  dc-local    global
server-2.global  172.16.7.102  4648  alive   false   2         0.6.0  dc-local    global
```

### notes 
1. Currently we don't preface the attribute keys with the tasker name -- I know @mcaulfield suggested this earlier, but feel like it might get cumbersome? Plus it's possible that down the road, two taskers might share the same attribute name like "MAC" or something. Prefacing with "kube:MAC" would prevent other taskers from using that attribute.
2. Right now, nomad tasker doesn't return an error if the attributes are not nomad-readable, and just doesn't make the constraint for that attribute-- do we want it to warn or alert? Feel like that could make sigma logs noisier, so holding off for now. 

### todo 
- Docs site: add Attributes to plugins and services doc 
- Docs site: create a nomad tasker how-to that explains how to add Constraints to services


manager/manager.go
```diff
@@ -593,15 +593,19 @@ func (m *manager) updateOperations(op model.ServiceOperation) {
 }
 
 func (m *manager) startOperation(op model.ServiceOperation) {
-	// get the site for this service op
-	site, err := m.db.GetSite(op.SiteRef.GetID().String())
+	allSites, err := m.gatherChildSites(op.SiteRef)
 	if err != nil {
+		log.Error(err)
+		m.updateOperationStatus(op.GetID(), model.ServiceOperationStatusFailed)
```
mikeynap: do we want to keep going here? or should we return?

manager/manager.go
```diff
@@ -880,34 +878,28 @@ func (m *manager) secretURLFromToken(t string) string {
...
 	}
-	// append self to the pool choose a child at random
-	children = append(children, parentSiteRef)
-	index := rand.Intn(len(children))
-	return children[index], nil
+	return children, nil
 }
 
-func (m *manager) gatherChildSitesWorker(originID, parentID string, isFirst bool) ([]model.ModelRef, error) {
+func (m *manager) gatherChildSitesWorker(originID, parentID string, isFirst bool) ([]model.Site, error) {
```
mikeynap: Do we have any loop detection in the sites logic? What if A sets B as a parent, and B sets A as a parent? Are there any checks in validation that would disallow that? Because if there is not, we should add a rudimentary loop detector here (i.e. make sure there are no dups).

task/docker.go
```diff
@@ -247,6 +252,12 @@ func (d *DockerTasker) read(ctx context.Context, reader io.ReadCloser, closer fu
 	}
 }
 
+// ChooseSite returns a site in sites at random
+func (d *DockerTasker) ChooseSite(sites []model.Site) (model.Site, error) {
+	index := rand.Intn(len(sites))
```
mikeynap: add a check here that len(sites) != 0

task/nomad.go
```diff
@@ -194,6 +201,56 @@ func (nt *NomadTasker) createPluginTask(so model.ServiceOperation) (*nomadapi.Ta
...
+		"${attr.cpu.numcores}", "${attr.cpu.totalcompute}", "${attr.consul.datacenter}",
+		"${attr.driver.<property>}", "${attr.unique.hostname}", "${attr.unique.network.ip-address}",
+		"${attr.kernel.name}", "${attr.kernel.version}", "${attr.platform.aws.ami-id}",
+		"${attr.platform.aws.instance-type}", "${attr.os.name}", "${attr.os.version}",
+	}
+
+	cons := []*nomadapi.Constraint{}
+	for k, v := range attr {
+		if stringInSlice(k, interpolatedVars) {
+			valMap := v.(map[string]interface{})
```
mikeynap: Makes me angsty... add an `valMap, ok` check (incase of malicious input)

task/nomad.go
```diff
@@ -194,6 +201,56 @@ func (nt *NomadTasker) createPluginTask(so model.ServiceOperation) (*nomadapi.Ta
...
+			if val == nil {
+				continue //constraint without value is invalid
+			}
+			op := valMap["operand"]
+			if op == nil {
+				op = "="
+			}
+			c := nomadapi.Constraint{
+				LTarget: k,
+				RTarget: val.(string),
```
mikeynap: we can't really know fo sho these are strings... I hate all the checks need, but maybe add the string conversion to the above nil checks.

Like `if _, ok := op.(string); !ok || op == nil { op = "=" } `

mikeynap: WLOG with `val`

task/nomad.go
```diff
@@ -207,3 +264,12 @@ func stringInSlice(a string, list []string) bool {
 func stringToPtr(str string) *string {
 	return &str
 }
+
+// AssembleAttributes combines the attributes maps from the serviceOp (key precedence) and Site
+func AssembleAttributes(so model.ServiceOperation, site model.Site) map[string]interface{} {
+	attr := site.Attributes
+	for k, v := range so.Attributes {
```
mikeynap: rewrite as: 
```
	for k, v := range so.Attributes {
                v := v
 		attr[k] = v
 	}
```

manager/manager.go
meganokeefe: probably return

PR #320, opened 2017-08-17
fixing nightly builds

stephenrlouie:


stephenrlouie:
fixed on crontab by adding a PATH variable in the cron file

Issue #315, opened 2017-08-15
Query objects needs id->name transformation 


meganokeefe:
Found the bug by using nullDB + trying to do an API query like 
`http://127.0.0.1:24601/v1/services?site=london` 

Which wasn't returning any objects despite that filter getting successfully passed to `QueryServices.`  Tested w/ couchbase and the same behavior happens. (was expecting 2 objects to be returned) 

DB Query functions need an id->name transformation and a siteRef query.

Issue #317, opened 2017-08-16
Couchbase shouldn't get by name 


meganokeefe:
Remove `idFromName` calls in GET 
https://github.com/sigma-dev/sigma/blob/master/database/couchbase.go#L313

PR #321, opened 2017-08-17
Consistent db

mikeynap:
- [x] Couchbase no longer supports getting an object by its name and id through the same function (either query with a name, or get directly with an id). If we want this capability, we should use the `database.GetModelRef(name,type,db)` or something similar (possibly augmented to return the object not just the id).
- [x] Database operations assume in queries that the objects in the database already have nil booleans defaulted to `false`
- [x] There are now `Database` integration tests which are reworked couchbase tests that have been reworked to test both databases and updated to cover more cases.
- [x] There is a `GenerateID()` so that the database can generate the appropriate ID for an object (secret IDs are different than other objects)
- [x] Consolidated the 'Name' fields in objects to exist only in the `Metadata` (reasoning for the `GenerateID()` function

Resolves: #315 #317


database/nulldb.go
```diff
@@ -232,6 +232,12 @@ func (n *NullDB) insertWithName(obj model.Metadataer, oType model.MetaType, objI
 	}
 	defer finish()
 
+	// make copy of metadata to restore
+	tmp := make(map[string]interface{})
```
adamschaub: What's the reason behind this save-and-restore? I'm confused about what this accomplishes.

tests/integration/database/database_test.go
```diff
@@ -0,0 +1,2554 @@
...
+				err:      nil,
+			},
+			{
+				query: func(base model.ServiceOperation) model.ServiceOperation {
+					casVal, _ := objs[0].Get(cas[name])
+					base.Set(cas[name], casVal)
+					return base
+				}(model.ServiceOperation{}),
+				expected: []model.ServiceOperation{objs[0]},
+				message:  "CAS only query",
```
adamschaub: I don't think we need to be checking queries for CAS. CAS is treated slightly differently between databases anyway (nulldb saves CAS with the object, while couchbase tracks CAS values separate from the object itself). Nobody should be attempting to query objects with a CAS value.

database/nulldb.go
vgunning: Couchbase doesn't mutate the object inserted, so it was screwing up the tests when the before and after objects didn't match (plugins etc suddenly had nulldb cas values after going through the insert operation even when a 'get' operation hadn't been performed).

tests/integration/database/database_test.go
vgunning: 👍 will remove those tests (they were already there)

vgunning: ...what would be expected behavior? don't allow? or just don't support a specific behavior

database/nulldb.go
```diff
@@ -232,23 +254,41 @@ func (n *NullDB) insertWithName(obj model.Metadataer, oType model.MetaType, objI
...
 
 	dat, err := json.Marshal(&obj)
 	n.objs[oType][id] = dat
 	n.nameToID[oType][objName] = id
 
+	// Restore old data
+	obj.SetID("")
+	obj.SetName("")
+	obj.Delete(model.SigmaMetaID)
+	obj.Delete(model.SigmaMetaName)
```
adamschaub: No need to strip ID/name. Just remove CAS.

database/nulldb.go
```diff
@@ -232,23 +254,41 @@ func (n *NullDB) insertWithName(obj model.Metadataer, oType model.MetaType, objI
...
+	obj.SetName(objName)
 	obj.Set(casKey, 1)
 
 	dat, err := json.Marshal(&obj)
 	n.objs[oType][id] = dat
 	n.nameToID[oType][objName] = id
 
+	// Restore old data
+	obj.SetID("")
+	obj.SetName("")
```
adamschaub: Redundant with 285-286?

database/nulldb.go
```diff
@@ -232,23 +254,41 @@ func (n *NullDB) insertWithName(obj model.Metadataer, oType model.MetaType, objI
 	}
 	defer finish()
 
-	id := model.GenerateMetaID()
+	// make copy of metadata to restore
+	tmp := make(map[string]interface{})
+	for key, val := range obj.GetMetadata() {
+		tmp[key] = val
+	}
```
adamschaub: What's the use of this?

model/metadata.go
```diff
@@ -28,6 +28,8 @@ type Metadataer interface {
 	// ID, Name are included in all objects
 	GetID() MetaID
 	SetID(MetaID)
+	GenerateID() MetaID
```
adamschaub: I don't think this belongs in this interface. Could just stay a single static generator. What's the reason for the change?

vgunning: I added it with the secrets stuff since the secrets generate ids differently. I wanted the database to be able to call the `obj.GenerateID()` to get a compatible id for whatever type it is.
I didn't look to deeply into the secret specifics, and just wanted to make it easy to interface with with the already existing implementation.

model/metadata.go
```diff
@@ -180,11 +182,18 @@ func (m *MetaImpl) Delete(k string) {
...
-		MetaID:   id,
-		MetaName: name,
-	}
+	mi := MetaImpl{}
 	mi.SetID(id)
 	mi.SetName(name)
 	return mi
 }
+
+func (m MetaImpl) Clone() MetaImpl {
```
adamschaub: Not used/remove

PR #322, opened 2017-08-17
Fixed Updates

### LINKED ISSUE #306: Update Endpoints not functioning correctly
It remains to be seen if the naming changes will correct this, but currently UPDATE /services,etc do not work as intended. 

An example: updating a service causes the previous vars to be deleted.

These endpoints could use some love to ensure everything up to snuff.

mikeynap:
Closes #306 
1. Updates Now Work for Services.
2. Added Update Tests to Functional Tests
3. Fixed GetServiceOperations() [ was previously not working due to trying to match the HeartbeatTTL -- fix was not being able to query on time.Durations.]
4. Fixed a _tragic_ one line error in convert.go -- optionally required services should NOT work on master at all right now.


api/v1/server/server.go
```diff
@@ -569,7 +569,7 @@ func (s *Server) AddServiceOperationOutput(w http.ResponseWriter, r *http.Reques
 		return
 	}
 
-	if _, ok := apiObj.Vars.(map[string]interface{}); !ok {
+	if _, ok := apiObj.Vars.(map[string]interface{}); !ok && apiObj.Vars != nil {
```
adamschaub: Do we want to accept output if it is nil? I reckoned it would at minimum be an empty map.

api/v1/server/server.go
```diff
@@ -1259,17 +1261,23 @@ func (s *Server) UpdateService(w http.ResponseWriter, r *http.Request) {
...
 		var err error
 		if service2Update, err = s.db.GetService(id); err != nil {
 			return http.StatusNotFound, err
 		}
 		updatedService.MetaImpl = service2Update.MetaImpl
+		updatedService.Attributes = service2Update.Attributes
 		updatedService.Status = service2Update.Status
 		updatedService.PreviousVars = service2Update.PreviousVars
-
+		for k, v := range updatedService.Vars { // Generated is not exposed on the API object, backfill it.
```
adamschaub: good catch

tests/functional/main_test.go
```diff
@@ -31,6 +31,7 @@ func TestMain(m *testing.M) {
 	defaultAddr := fmt.Sprintf("%s:%d", defaultIP, 24816)
 
 	flag.StringVar(&sigtest.Config.APIURL, "address", defaultAddr, "sigma api endpoint address")
+	flag.BoolVar(&sigtest.Config.SleepMode, "sleep", false, "use test setup, but sleep instead of running the tests.")
```
adamschaub: what

tests/functional/sigtest/manager.go
```diff
@@ -169,6 +171,7 @@ func testServices(t *testing.T) {
 	for _, svc := range services.getNotPassed() {
 		if shouldPass(svc) && svc.Status != model.ServiceStatusActive {
 			t.Errorf("Service %s: Status was %s, should have been active.", svc.Name, svc.Status)
+			fmt.Printf("Service %s: Status was %s, should have been active.", svc.Name, svc.Status)
```
adamschaub: remove

tests/functional/sigtest/manager.go
```diff
@@ -184,33 +187,48 @@ func testServices(t *testing.T) {
 // 3) They become active again.
 func testServicesUpdate(t *testing.T) {
 	type updateTest struct {
-		svc     string
-		varName string
-		val     interface{}
+		svc  string
+		vars []interface{}
+		vals []interface{}
```
adamschaub: same here

tests/functional/sigtest/run.go
```diff
@@ -169,7 +173,7 @@ func setupCommon(cfg *TestConfig) {
 	}
 
 	cfg.ManagerOperationUpdateInterval = "200ms"
-	cfg.ManagerServiceStatusInterval = "3s"
+	cfg.ManagerServiceStatusInterval = "300s"
```
adamschaub: why so large?

tests/functional/sigtest/manager.go
```diff
@@ -184,33 +187,48 @@ func testServices(t *testing.T) {
 // 3) They become active again.
 func testServicesUpdate(t *testing.T) {
 	type updateTest struct {
-		svc     string
-		varName string
-		val     interface{}
+		svc  string
+		vars []interface{}
```
mikeynap: yeah originally I was gonna do updates on non-strings, but figured there was no point. I'll change it

tests/functional/sigtest/run.go
mikeynap: I didn't want statuses to happen when I'm doing updates (if both ran at the same time, status reporting could over write updates)

tests/functional/main_test.go
mikeynap: don't worry about it lol

PR #348, opened 2017-09-13
Plugin/Service Version Reqs Fix

### LINKED ISSUE #342: Plugin version weirdness
When a plugin requires another plugin and they're different versions, sigma won't deploy the plugin. 

need to look into.

vgunning:
### PR checklist

### Change log

Added in manager tests to make sure that `getReqServices` was behaving properly.
Fixed `getReqServices` so that it is comparing the dependencies and their plugin versions to ensure that they match the given service's dependencies.

Resolves #342


manager/manager.go
```diff
@@ -374,37 +376,39 @@ func (m *manager) removeService(service model.Service) {
...
-		svc, err = m.db.GetService(req.GetID().String())
+		var svcDep model.Service
+		svcDep, err = m.db.GetService(req.GetID().String())
 		if err != nil {
 			return
 		}
+
 		// Update ID
-		service.Requirements[idx].SetID(svc.GetID().String())
+		service.Requirements[idx].SetID(svcDep.GetID().String())
```
meganokeefe: worth checking for a nil `svcDep` here?

manager/manager.go
```diff
@@ -374,37 +376,39 @@ func (m *manager) removeService(service model.Service) {
...
 			return
 		}
 		for i, pluginReq := range plugin.Requirements {
-			if pluginReq.Name == pluginRefObj.GetName() {
+			// If the plugin is listed as a req, check that it fits the constraints
+			if pluginDep.GetName() == pluginReq.Name {
 				// Versions and constraints are validated via the API, no need to catch
 				// errors here
-				vers, _ := version.NewVersion(plugin.Version)
+				vers, _ := version.NewVersion(pluginDep.Version)
```
meganokeefe: nice catch.

PR #359, opened 2017-09-19
Add progress description and completion to service operation

adamschaub:
Add description + completion to heartbeat. Keep track in service operation.


api/v1/gen/v1.yaml
```diff
@@ -658,10 +664,20 @@ definitions:
...
+      completion:
+        type: integer
       output:
+        $ref: '#/definitions/ServiceOperationOutput'
+      vars:
         type: object
-      input:
+        additionalProperties:
+          $ref: '#/definitions/ServiceVar'
+      prevoius-vars:
```
rfliegerallison: Small typo here

adamschaub: It's French

PR #363, opened 2017-09-20
Add functionality for REST API requests by name or ID

### LINKED ISSUE #357: Sigma API REST API by name or id for all endpoints
##### Description

Today Sigma API only allows GET/POST/DELETE by "id". Allow users to alternatively specify "name" in place of "id" field (in the URL) as long as the name is unique.

##### Sigma Version

0.3

##### Sigma Configuration

NA

##### Command line used for generation / environment

For example, this should work:
`./api/v1/cli/sigma-cli getService --host "localhost:24601" id="service-name"`

##### Steps to reproduce

NA

##### Related issues

NA

##### Suggest a Fix

Update the code in `./api/v1/server/server.go` and `./database.go` for `getModelRef`.

rfliegerallison:
### PR checklist

### Change log

Adds brief segments to each of the get/delete/update/reset functions for plugins, services, and sites in `server.go`. Also, tests have been added to `server_test.go` to assert correct behavior in response to valid and invalid use of by-name API requests.

### Notes

Resolves #357


api/v1/server/server_test.go
```diff
@@ -493,6 +498,66 @@ func TestDeletePlugin(t *testing.T) {
...
+func TestDeletePluginByName(t *testing.T) {
+
+	// Define table-driven tests.
+	var tests = []struct {
+		pluginsToAdd             []apiclient.Plugin
+		pluginToDelete           apiclient.Plugin
+		expectedErrCode          int
+		expectedRemainingPlugins []apiclient.Plugin
+	}{
+		{ // Deleting valid plugin by name.
```
vgunning: A common thing in tests is to have them print out a message when a particular test fails, that way we know which particular test failed in the `TestFunction`
Such as https://github.com/sigma-dev/sigma/blob/master/tests/integration/database/database_test.go#L107

api/v1/server/server_test.go
```diff
@@ -493,6 +498,66 @@ func TestDeletePlugin(t *testing.T) {
 	}
 }
 
+func TestDeletePluginByName(t *testing.T) {
+
```
vgunning: Could add `defer db.Clear()` to the beginning of the function in case it hits too many errors or fails to completely run for some reason.

api/v1/server/server_test.go
```diff
@@ -493,6 +498,66 @@ func TestDeletePlugin(t *testing.T) {
...
+			t.Errorf("Expected %d, got %d", test.expectedErrCode, res.StatusCode)
+		}
+
+		// Assert that the remaining data is what is expected.
+		if val, _, _ := client.GetPluginsAll(); !apiPluginListsEqual(test.expectedRemainingPlugins, val) {
+			t.Errorf("Remaining plugins don't match expected")
+		}
+
+		// Reset the database after each test.
+		db.Clear()
```
vgunning: Clearing before you start the loop/add to the db, may be 'safer'

api/v1/server/server_test.go
```diff
@@ -649,6 +920,90 @@ func TestGetService(t *testing.T) {
...
+	}
+
+	// Run the tests.
+	for _, test := range tests {
+
+		// Add sites.
+		for _, site := range test.sitesToAdd {
+			_, err := db.InsertSite(convert.FromAPISite(site, db))
+			if err != nil {
+				t.Errorf("Unexpected error: %v", err)
```
vgunning: Could be more verbose on what is throwing an error (i.e. you have three different places where you could pass back an `Unexpected error:`

api/v1/server/server_test.go
```diff
@@ -743,6 +1224,100 @@ func TestUpdateService(t *testing.T) {
...
+		// Update the service by name.
+		res, _ := client.UpdateService(test.serviceToUpdate.Name, test.updatedService)
+
+		// Assert correct response codes.
+		if res.StatusCode != test.expectedErrCode {
+			t.Errorf("Expected %d, got %d", test.expectedErrCode, res.StatusCode)
+		}
+
+		// Assert that the remaining data is what is expected.
+		if val, _, _ := client.GetServicesAll(); !apiServiceListsEqual(test.expectedServicesAfter, val) {
```
vgunning: Instead of relying on the `client.Get<Type>All()` you could query the db directly.

api/v1/server/server_test.go
```diff
@@ -848,28 +1539,100 @@ func marshalOrDie(t *testing.T, v interface{}) string {
...
+
+func apiServicesEqual(s1, s2 apiclient.Service) bool {
+	s1.Id = ""
+	s2.Id = ""
+	s1.LastUpdated = time.Time{}
+	s2.LastUpdated = time.Time{}
+	return reflect.DeepEqual(s1, s2)
+}
+
+func apiSiteListsEqual(site1, site2 []apiclient.Site) bool {
```
vgunning: consistency NIT, plural? `sites` instead of `site`

api/v1/server/server_test.go
```diff
@@ -910,6 +1673,23 @@ var fooPlugin = apiclient.Plugin{
 	},
 	Container: "foo:1.0",
 }
+var fooPluginUpdated = apiclient.Plugin{
+	Id:      "new_foo",
+	Name:    "new_foo",
```
vgunning: Is this supposed to be mimicking a new version of the 'foo' plugin? Otherwise, since this isn't actually testing updating a plugin object... could it be named something less ambiguous on its purpose
If so, probably only the version should be updated, and the service should take the new `foo:2.0` or whatever

rfliegerallison: The issue here is that when performing an update on a service and change the plugin it's using, in order for the update to be "valid", the plugin you want to change to has to have already been added to the registry of plugins. BUT you can't add two plugins with the same ID, which is why I made them distinctly different. The only reason I kept the name fooPlugin is because I didn't alter any of the Vars

api/v1/server/server_test.go
rfliegerallison: It seems too complicated, since the database doesn't offer any way to get ALL plugins, services, etc. So even though I can check that all the expected plugins/services/sites are in the database as expected, I can't see a clear way to check that nothing ELSE is also in the database that I don't want to be there.

api/v1/server/server_test.go
```diff
@@ -493,6 +498,73 @@ func TestDeletePlugin(t *testing.T) {
...
+		// Insert plugins.
+		for _, plugin := range test.pluginsToAdd {
+			_, err := db.InsertPlugin(convert.FromAPIPlugin(plugin))
+			if err != nil {
+				t.Errorf("Unexpected error while adding plugins: %v (%s)", err, test.description)
+			}
+		}
+
+		// Delete a plugin by name:version.
+		res, _ := client.DeletePlugin(test.pluginToDelete.Name + ":" + test.pluginToDelete.Version)
```
vgunning: In case our format for the whole plugin tag changes, there's the `model.ToPluginNameVer(name, version)`

PR #366, opened 2017-09-26
GoodbyeWorld example fix 

meganokeefe:
Unmarshalling required variables and sending statuses was outdated/bad.


examples/goodbye-world/main.go
```diff
@@ -147,7 +154,7 @@ func (gw *GoodbyeWorld) writeOutputs(op string) error {
 	gw.gVars.Output = fmt.Sprintf("%s: Goodbye %s", op, spl[1])
 	gw.gVars.Number = rangeIn(1000, 9999) // generate a random 4 digit number
 	fmt.Printf("Writing [%s] and [%d] to vars", gw.gVars.Output, gw.gVars.Number)
-	return gw.client.SendOutput(gw.gVars, "COMPLETE")
+	return gw.client.SendOutput(gw.gVars, "Succeeded")
```
stephenrlouie: lets make the statuses an exported CONST in plugin_client

plugin/client/plugin.go
```diff
@@ -22,6 +22,8 @@ const (
 	PluginArgStatus              = "status"
 	PluginArgUpdate              = "update"
 	PluginArgUndefined           = "undefined"
+	StatusSucceeded              = "Succeeded"
```
stephenrlouie: nit. ref https://github.com/sigma-dev/sigma/blob/d12f38c9ee3569a80daa4ef5d0b691f78c8a0fc8/model/model.go#L172-L173

PR #370, opened 2017-09-27
Adds assertion that services marked for removal cannot be updated

### LINKED ISSUE #367: Prevent services from updating once marked for deletion
Once a user calls `DeleteService()`, further updates to that service should be prevented by the API. 

This makes it so that a user can't "reverse" a deletion by setting the `Remove` flag back to `false.` And reduces ambiguity.

rfliegerallison:
### PR checklist

### Change log

Adds a one-line assertion in UpdateService to make sure the service hasn't already been removed. Also adds unit test.

### Notes

Resolves #367


api/v1/server/server.go
```diff
@@ -1347,6 +1348,8 @@ func (s *Server) UpdateService(w http.ResponseWriter, r *http.Request) {
 		var err error
 		if service2Update, err = s.db.GetService(id); err != nil {
 			return http.StatusNotFound, err
+		} else if service2Update.Remove != nil && *(service2Update.Remove) {
```
meganokeefe: small nit, don't think you need the else here, just if!

PR #375, opened 2017-09-28
Adds `self` feature to examples and reorganizes example scripts

### LINKED ISSUE #371: Examples to use self
##### Description

After merging #368, our examples should also use self to parse their own inputs. 

For example: https://github.com/sigma-dev/sigma/blob/goodbyefix/examples/goodbye-world/main.go#L67 We can use the self.Name here instead

rfliegerallison:
### PR checklist

### Change log

Removes old `/examples` scripts and add/updates scripts to be more refactored and easier to run.

Also, I updated both examples' `main.go` files to use `self` now that it's been added as a feature.

### Notes

Resolves #371 

### TODO

I still need to update the README to reflect the changes I've made, if they seem reasonable.


examples/README.md
```diff
@@ -6,7 +6,7 @@
 
 #### Optional
 - Running Couchbase container (if you do not have this, run Sigma with `--nulldb`)
```
meganokeefe: can you change this to `--use-nulldb`?

examples/scripts/add-cali-sites.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
```
meganokeefe: i think you can probably remove everything except `add-site.sh`

rfliegerallison: okay will do. i wasn't sure if anyone was used to using the other scripts

PR #377, opened 2017-09-28
Controlled Updates in a dependency tree

stephenrlouie:
### Change log

- The update logic will now look at all of its requirements and ensure their `previousVars == Vars`, this being true that will imply that the requirement has gone through its update. It `previousVars != Vars` an update still needs to happen. We do not recursively go down the dependency tree.

### Notes
It is possible for more updates to happen since this varsCheck will only go one level deep into requirements. If we have services A <- B <- C. Updating C and A at the same time. C will see that B is not in flux and update. A will update, causing B to updating which will again cause C to update. Eventual consistency here.

### TODO

- [x] fix integration tests, ditto plugin is a mess
- [x] leave notes behind on ditto and update tests


config.yaml
```diff
@@ -1,11 +1,11 @@
 ---
```
stephenrlouie: i will remove before we merge

manager/manager.go
```diff
@@ -421,6 +418,53 @@ func (m *manager) getReqServices(service model.Service, plugin model.Plugin) (ou
 	return
 }
 
+func (m *manager) getOptServices(service model.Service, plugin model.Plugin) (out []model.Service, err error) {
```
stephenrlouie: not in use right now

api/v1/server/server.go
```diff
@@ -1377,13 +1377,17 @@ func (s *Server) UpdateService(w http.ResponseWriter, r *http.Request) {
 		}
 
 		updatedService.MetaImpl = service2Update.MetaImpl
-		updatedService.Status = service2Update.Status
 		updatedService.PreviousVars = service2Update.PreviousVars
 		for k, v := range updatedService.Vars { // Generated is not exposed on the API object, backfill it.
 			v := v
 			if oldV, ok := service2Update.Vars[k]; ok {
 				v.Generated = oldV.Generated
```
adamschaub: `v` is a local var in this context, so `v.Generated` will not be propagated in any sense. Maybe looking for `updatedService.Vars[k].Generated`?

manager/manager.go
```diff
@@ -421,6 +418,53 @@ func (m *manager) getReqServices(service model.Service, plugin model.Plugin) (ou
...
+	optsMet := make([]bool, len(plugin.Optional))
+	for idx, opt := range service.Optional {
+		var svcDep model.Service
+		svcDep, err = m.db.GetService(opt.GetID().String())
+		if err != nil {
+			return
+		}
+
+		// Update ID
+		service.Optional[idx].SetID(svcDep.GetID().String())
```
adamschaub: I don't get why this is necessary. We just found this service by using `opt.GetID()`.

manager/manager.go
```diff
@@ -474,6 +518,25 @@ func serviceAssetName(id, vName string) string {
 	return fmt.Sprintf("%s.%s", id, vName)
 }
 
+func (m *manager) isReadyForUpdate(reqs []model.Service) bool {
```
adamschaub: Doesn't apply for a reqs of reqs. Have to recurse all the way up all reqs.

api/v1/server/server.go
stephenrlouie: oops, good catch.

stephenrlouie: I just want `oldV.Generated`. updated

manager/manager.go
stephenrlouie: TBH this was copy pasta from `GetReqServices()`

manager/manager.go
stephenrlouie: it does not, I figured we can keep it simple for now. This may cause updating multiple times. But that shouldn't be harmful. I don't know how long we want to block an update just because a service recursively far away will be updating. Talked to matt on this, we think this is fine. We might not even have to block at all now that vars are more tightly controlled and outputs cannot disappear from an update.

manager/manager_test.go
```diff
@@ -503,26 +507,45 @@ func TestUpdateService(t *testing.T) {
...
-		//				// Should trigger input changed
-		//				Input: map[string]model.ServiceDataItem{
-		//					"": model.ServiceDataItem{Value: []string{"hello", "world"}, DataItem: model.DataItem{Type: "list/string"}},
-		//				},
-		//				Disabled: model.NewBool(false),
-		//			},
-		//			plugin:       model.Plugin{},
-		//			inputChanged: true,
-		//		},
+		// {
```
meganokeefe: remove this?

tests/functional/ditto/ditto2-plugin.json
```diff
@@ -0,0 +1,21 @@
+{
+	"name": "ditto2",
```
meganokeefe: nit can all the things named `ditto2` be renamed to just `ditto`?

PR #388, opened 2017-10-09
Updated example plugins to not clobber Deploy() variables

meganokeefe:
This was my fault -- when I wrote the new example plugins (and goodbye, which I wrote weeks beforehand), I was having `Status()` and `Update()` calls return blank variables to sigma, which overwrites the result of deploy.


examples/coffee-run/coffee-orders/main.go
```diff
@@ -87,15 +87,21 @@ func (c *CoffeeOrders) destroy(op pc.PluginArg) error {
...
+
 	if err := c.writeOutputs(true); err != nil {
 		return err
 	}
 	return nil
 }
 
 func (c *CoffeeOrders) update(op pc.PluginArg) error {
-	fmt.Println("CoffeeOrders Update")
+	fmt.Println("CoffeeOrders Deploy")
```
mikeynap: don't you mean update

adamschaub: It's deploying the update

examples/coffee-run/coffee-prices/main.go
```diff
@@ -114,23 +114,27 @@ func (c *CoffeePrices) destroy(op pc.PluginArg) error {
 }
 
 func (c *CoffeePrices) status(op pc.PluginArg) error {
-	fmt.Println("CoffeePrices Status")
+	fmt.Println("CoffeePrices Deploy")
```
adamschaub: Status

examples/coffee-run/coffee-prices/main.go
```diff
@@ -114,23 +114,27 @@ func (c *CoffeePrices) destroy(op pc.PluginArg) error {
...
+	fmt.Printf("Generated coffee prices: %v, TOTAL: %v", c.Vars.CoffeePrices, c.Vars.Total)
 	if err := c.writeOutputs(true); err != nil {
 		return err
 	}
 	return nil
 }
 
 func (c *CoffeePrices) update(op pc.PluginArg) error {
-	fmt.Println("CoffeePrices Update")
+	fmt.Println("CoffeePrices Deploy")
```
adamschaub: Update

PR #391, opened 2017-10-12
Removes parent in sites

stephenrlouie:


meganokeefe:
LGTM on a first pass, will use now for testing...

PR #395, opened 2017-10-16
Service Labels and Selectors

### LINKED ISSUE #351: Sigma Labels/Selectors
Allow Sigma services to select service dependencies via labels/selectors in addition to explicit service names (i.e. k8s service can just grab all node services in a site and whenever that list changes, an update is triggered w/o explicitly adding/removing that service from the k8s service definition).

meganokeefe:
Closes #351 

### changelog 

- [x] model.go / API -- plugins now only have requirements, not optional. services have a set of string labels, and requirements are still a list of strings (but are now selectors instead of a list of explicit required services). services have no optional services. finally, pluginreqs now have the `min` field, specifying the minimum cardinality of that required service to get the other service to run. 
- [x] API convert 
- [x] validation for selectors and labels -- ensuring strings are of the right format
- [x] manager logic -- resolving selectors to explicit lists of services + checking cardinality 
- [x] updated unit tests 
- [x] updated `examples/` [with four new plugins that use different Mins](https://github.com/sigma-dev/sigma/tree/labels-selectors/examples/selectors)
- [x] update existing examples to use the new `requirements` format. 
- [x] [Docs PR](https://github.com/sigma-dev/sigma/pull/397)

### how to test 

- go into `./selectors/examples` and run `./load.sh` which sets up a bunch of services using labels + selectors. Can watch them run in the expected order, etc. 
- you can customize the services in that folder to experiment with more complex labels and selectors 
- you can also update any of the other examples that use requirements (goodbye-world, my-weather, coffee-process) to use labels. Right now the selectors in those services are very simple, ie. `service=hello-world-svc`.


api/v1/gen/index.html
```diff
@@ -184,7 +184,7 @@
 </div>
     <div class="app-desc">More information: <a href="https://helloreverb.com">https://helloreverb.com</a></div>
     <div class="app-desc">Contact Info: <a href="hello@helloreverb.com">hello@helloreverb.com</a></div>
-    <div class="app-desc">Version: 1.0</div>
+    <div class="app-desc">Version: 3.0</div>
```
adamschaub: 1 -> 3?

api/v1/server/server.go
```diff
@@ -973,6 +973,7 @@ func (s *Server) GetPlugins(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	obj := model.Plugin{}
+	log.Warnf("Filters: %#v", filters)
```
adamschaub: Remove before merge

api/v1/server/validate.go
```diff
@@ -17,9 +17,13 @@ import (
 )
 
 var (
-	ErrNoServiceSite     = errors.New("No Site provided for Service")
-	ErrSiteNotFound      = errors.New("Site provided not found")
-	ErrInvalidSiteFormat = errors.New("Invalid site format, cannot contain more than one semicolon")
+	ErrNoServiceSite       = errors.New("No Site provided for Service")
+	ErrSiteNotFound        = errors.New("Site provided not found")
+	ErrInvalidSiteFormat   = errors.New("Invalid site format, cannot contain more than one semicolon")
+	ErrEmptyLabel          = errors.New("Service label cannot be an empty string")
```
adamschaub: Can it not? I don't see why empty string can't be a viable value. 

(Still have to read the implementation, so I might delete this comment if I see I'm wrong)

api/v1/server/validate.go
```diff
@@ -17,9 +17,13 @@ import (
...
+	ErrReservedLabel       = errors.New("Label key cannot be one of: site, service, plugin")
```
adamschaub: Plugin version?

api/v1/server/validate.go
```diff
@@ -169,23 +176,46 @@ func (s *Server) validateService(service *apiclient.Service, isUpdate bool) erro
...
 	}
 
-	for _, req := range service.Requirements {
-		svcRef := database.GetModelRef(req, model.ServiceType, s.db)
-		if len(svcRef) == 0 {
-			return fmt.Errorf("No existing service with name %s", req)
+	// Validate labels - no empty strings allowed, no reserved labels
+	for k, _ := range service.Labels {
+		if len(k) == 0 {
+			return ErrEmptyLabel
```
adamschaub: Important?

api/v1/server/validate.go
```diff
@@ -169,23 +176,46 @@ func (s *Server) validateService(service *apiclient.Service, isUpdate bool) erro
...
 }
 
+// returns true iff the selector is formatted correctly
+/* examples:
+A=x&B=y (labels)
+service=helloSvc&site=central
+service=helloSvc
+*/
+func IsValidSelector(sel string) bool {
+	//TODO - possibly expand these operators later
```
adamschaub: IRRELEVANT. Capture in issue

api/v1/server/validate.go
```diff
@@ -169,23 +176,46 @@ func (s *Server) validateService(service *apiclient.Service, isUpdate bool) erro
...
+	pairs := strings.Split(sel, "&")
+	for _, p := range pairs {
+		elts := strings.Split(p, "=")
+		if len(elts) != 2 {
```
adamschaub: FWIW, if `p == "label="`, the `len(elts)` would still be 2. Shouldn't offer any difficulties.

api/v1/server/validate.go
```diff
@@ -169,23 +176,46 @@ func (s *Server) validateService(service *apiclient.Service, isUpdate bool) erro
...
+			return false
+		}
+		if strings.Contains(p, " ") { //no whitespace
```
adamschaub: Why no whitespace? Given the spec, it doesn't affect any of the delimiters. 

Either case, maybe add check for \t \n etc.

config.yaml
```diff
@@ -2,10 +2,11 @@
 Log:
   Level: info
 Manager:
+  UpdateInterval: 5s
   OperationUpdateInterval: 5s
   ServiceStatusInterval: 20s
 Docker:
-  APIUrl: 127.0.0.1:24601
+  APIUrl: 10.131.33.191:24601
```
adamschaub: rollback

config.yaml
```diff
@@ -21,5 +22,5 @@ Rabbit:
   Password: guest
   URL: localhost:5672
 Server:
-  Host: 127.0.0.1:24601
+  Host: 10.131.33.191:24601
```
adamschaub: rollback

examples/selectors/services/serviceA1.json
```diff
@@ -0,0 +1,8 @@
+{
+	"name": "a1",
+	"plugin": "example-a:0.1",
+	"site": "central",
+  "labels": {
```
adamschaub: indent gore

model/model.go
```diff
@@ -98,11 +98,18 @@ const (
 	ServiceStatusFailed   = "Failed"
 )
 
+var (
+	PluginSelector  = "plugin"
+	ServiceSelector = "service"
+	SiteSelector    = "site"
+)
```
adamschaub: PluginVersionSelector??

manager/manager.go
```diff
@@ -1012,3 +898,184 @@ func (s opSlice) Less(i, j int) bool {
...
+	}
+	return output, nil
+}
+
+// returns true iff cardinality is reached on all of this plugin's requirements
+// NOTE - we DO keep services that aren't in the requirements but were dredged up by a selector
+func (m *manager) allMinReached(plugin model.Plugin, serviceMap map[string]int) bool {
+	for _, req := range plugin.Requirements {
+		if req.Min > 0 { //where Min = 0 (Default), the requirement is "optional" and we don't care if it's in the map.
+			k := model.ToPlugNameVer(req.Name, req.Version)
```
adamschaub: req.Version might not be a single version. Could represent a range of versions

manager/manager.go
```diff
@@ -1012,3 +898,184 @@ func (s opSlice) Less(i, j int) bool {
...
+	return true
+}
+
+// resolveSelector resolves one Requirement selector string, and returns a set of service IDs
+func (m *manager) resolveSelector(selector string) (*sets.Set, error) {
+	kvPairs := strings.Split(selector, "&")
+	if len(kvPairs) < 1 {
+		return sets.New(), fmt.Errorf("Malformatted selector: %s", selector)
+	}
+	intersection, err := m.findServicesForPair(kvPairs[0]) //returns a list of svc IDs
```
adamschaub: this seems very wasteful. Is it not possible to do a query that simultaneously selects services with all labels?

manager/manager.go
```diff
@@ -1012,3 +898,184 @@ func (s opSlice) Less(i, j int) bool {
...
+		return output, err
+	}
+	// Check if all requirements' Min is met
+	minReached := m.allMinReached(plugin, minMap)
+	if !minReached {
+		return output, fmt.Errorf("Service %s has not reached min svc requirements", service.GetName())
+	}
+
+	// If cardinality is reached, return the []model.Service for the final union ID set
+	output, err = m.setToServiceSlice(union)
```
adamschaub: Redundant Service GETs between this and `setToMinMap`. Could do this first, and refactor `setToMinMap` to accept a slice of Services instead of redoing the lookup

manager/manager.go
```diff
@@ -1012,3 +898,184 @@ func (s opSlice) Less(i, j int) bool {
...
+	output := sets.New()
+	elts := strings.Split(pair, "=")
+	if len(elts) != 2 {
+		return output, fmt.Errorf("Selector %s is invalid, must be <key>=<value>", pair)
+	}
+	key := elts[0]
+	value := elts[1]
+	queryService := model.Service{}
+
+	// Construct a DB query based on selector pair: one of service, plugin, site, or label.
```
adamschaub: can we do this for all pairs simultaneously

manager/manager.go
```diff
@@ -1012,3 +898,184 @@ func (s opSlice) Less(i, j int) bool {
...
+		output.Add(q.GetID().String())
+	}
+	return output, nil
+}
+
+// converts a set of service IDs to a map from plugin name:version -> # of services for it.
+func (m *manager) setToMinMap(idSet *sets.Set) (map[string]int, error) {
+	output := map[string]int{}
+	asSlice := sets.StringSlice(idSet)
+	for _, id := range asSlice {
```
adamschaub: this is essentially repeating `setToServiceSlice`

manager/manager.go
meganokeefe: HA. yes. there is absolutely a way to do that. 

i am bonobo

api/v1/server/validate.go
meganokeefe: added support for version like so- you now say `plugin=example:1.0` instead of just `plugin=example`

api/v1/gen/index.html
meganokeefe: It turns out that `Version` is not the API version but the OpenJson Spec version - we had been using a version that was years old. @rfliegerallison's discovery.

meganokeefe: (This change isn't in my PR but clearly no one had `gen html`'ed in a while)

api/v1/server/validate.go
meganokeefe: yeah sure

api/v1/server/validate.go
meganokeefe: Yes i think so

model/model.go
meganokeefe: now part of PluginSelector, embedded in the value, eg. `"plugin=hello:0.1"`. 

I am personally reluctant to have a separate selector for PluginVersion and not sure what the valuable use case is in being able to dredge up all plugins of a specific version. 

When if you wanted to do that  ^ ie. create a grouping of plugins, you'd give them all a label.

I am also Pro the way I do `plugin=hello:0.1` as a combo name:version because it obeys our Service:Plugin syntax and doesn't introduce further complexity for the user.

manager/manager.go
meganokeefe: I forgot that sigma supported this feature -- do you mean that `req.Version` could be something like `>1.0`? 

would need to figure out how that would translate here

api/v1/server/validate.go
```diff
@@ -155,23 +162,46 @@ func (s *Server) validateService(service *apiclient.Service, isUpdate bool) erro
...
 	return nil
 }
 
+// returns true iff the selector is formatted correctly
+/* examples:
+A=x&B=y (labels)
+service=helloSvc&site=central
+service=helloSvc
+*/
+func IsValidSelector(sel string) bool {
```
stephenrlouie: why not include the reserved labels here? Unless this is called by others that will be using reserved labels?

config.yaml
```diff
@@ -2,10 +2,11 @@
 Log:
   Level: info
 Manager:
+  UpdateInterval: 5s
   OperationUpdateInterval: 5s
   ServiceStatusInterval: 20s
 Docker:
-  APIUrl: 127.0.0.1:24601
+  APIUrl: 10.131.32.83:24601
```
stephenrlouie: make local host

config.yaml
```diff
@@ -21,5 +22,5 @@ Rabbit:
   Password: guest
   URL: localhost:5672
 Server:
-  Host: 127.0.0.1:24601
+  Host: 10.131.32.83:24601
```
stephenrlouie: localhost

examples/selectors/containers/exampleA/Makefile
```diff
@@ -0,0 +1,17 @@
...
+IMAGE ?= bin/$(EXECUTABLE)
+
+all: build
+
+build:
+	CGO_ENABLED=0 go build --ldflags '${EXTLDFLAGS}' -o ${IMAGE} github.com/sigma-dev/sigma/examples/selectors/plugins/exampleA
+
+fast:
+	mkdir -p bin
+	GOBIN=$(GOPATH)/src/github.com/sigma-dev/sigma/examples/selectors/plugins/exampleA//bin go install --ldflags '${EXTLDFLAGS}' github.com/sigma-dev/sigma/examples/exampleA/
```
stephenrlouie: double `//`

examples/selectors/containers/exampleA/main.go
```diff
@@ -0,0 +1,124 @@
...
+	fmt.Println("Status...")
+	if err := ex.writeOutputs(); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (ex *ExampleA) update(op pc.PluginArg) error {
+	fmt.Println("Update...")
+	if err := ex.writeOutputs(); err != nil {
```
stephenrlouie: just a high level comment about plugin outputs. I forget if we are doing a complete overwrite anytime we receive output. This may cause us to accidentally remove values that Deploy had. Lets say deploy returns `certs` but Status just checks the endpoint and will not include `certs` in its `writeOutputs` this "may" cause the certs value to be lost.

examples/selectors/containers/exampleC/Makefile
```diff
@@ -0,0 +1,17 @@
...
+IMAGE ?= bin/$(EXECUTABLE)
+
+all: build
+
+build:
+	CGO_ENABLED=0 go build --ldflags '${EXTLDFLAGS}' -o ${IMAGE} github.com/sigma-dev/sigma/examples/selectors/plugins/exampleC
+
+fast:
+	mkdir -p bin
+	GOBIN=$(GOPATH)/src/github.com/sigma-dev/sigma/examples/selectors/plugins/exampleC//bin go install --ldflags '${EXTLDFLAGS}' github.com/sigma-dev/sigma/examples/exampleC/
```
stephenrlouie: double slash again

examples/selectors/containers/exampleD/Makefile
```diff
@@ -0,0 +1,17 @@
...
+IMAGE ?= bin/$(EXECUTABLE)
+
+all: build
+
+build:
+	CGO_ENABLED=0 go build --ldflags '${EXTLDFLAGS}' -o ${IMAGE} github.com/sigma-dev/sigma/examples/selectors/plugins/exampleD
+
+fast:
+	mkdir -p bin
+	GOBIN=$(GOPATH)/src/github.com/sigma-dev/sigma/examples/selectors/plugins/exampleD//bin go install --ldflags '${EXTLDFLAGS}' github.com/sigma-dev/sigma/examples/exampleD/
```
stephenrlouie: double slash

manager/manager.go
```diff
@@ -968,3 +851,156 @@ func (s opSlice) Less(i, j int) bool {
...
 }
+
+// getReqServices is responsible for resolving service Requirements selectors to concrete Sigma services.
+// returns an error if unable to resolve any of the selectors OR if cardinality is not reached for one or more plugin reqs
+func (m *manager) getReqServices(service model.Service, plugin model.Plugin) (output []model.Service, err error) {
+	if len(service.Requirements) == 0 {
+		return output, nil
+	}
+
+	// "union" is a combined pool of services from each selector. (takes care of potential duplicates)
```
stephenrlouie: why not include this in the loop and just make a `*sets.Set`

api/v1/server/validate.go
meganokeefe: yes ^ like all selectors, eg.` x=y, plugin=` ... call this

examples/selectors/containers/exampleA/main.go
meganokeefe: i'm gonna put out an issue for this

manager/manager.go
meganokeefe: fixed

PR #402, opened 2017-10-24
Selectors Bug fix + Func Tests Update 

meganokeefe:
1) Updated the test Plugins and Services in `tests/functional/sigtest/plugins_services_objects` to use the Req/Min format instead of Req/Opt.My bad for not catching this before Selectors merged.

2) Fixed an additional bug involving version checking on `allMinReached`, within manager. I was doing strict version equals instead of using the `version` package. I was also assuming (incorrectly) that all PluginReqs include a version. Addressed both issues.


manager/manager.go
```diff
@@ -885,21 +889,80 @@ func (m *manager) getReqServices(service model.Service, plugin model.Plugin) (ou
 	return output, nil
 }
 
+func (m *manager) serviceMapFromSlice(svcSlice []model.Service) (map[string][]model.Service, error) {
+	// Construct a temp data structure to make it easy to check reqs
+	serviceMap := make(map[string][]model.Service)
+	for _, svc := range svcSlice {
+		p, err := m.db.GetPlugin(svc.PluginRef.GetID().String())
```
adamschaub: Why are we getting plugin here? Services should already have plugin name/version

manager/manager.go
```diff
@@ -885,21 +889,80 @@ func (m *manager) getReqServices(service model.Service, plugin model.Plugin) (ou
...
+		if req.Min < 1 { //we don't care about Min=0 services
+			continue
+		}
+		nameConstraint := req.Name
+		var versionConstraint version.Constraints
+		var err error
+		if req.Version != "" {
+			versionConstraint, err = version.NewConstraint(req.Version)
+			if err != nil {
+				log.Error(err)
```
adamschaub: Possibly unnecessary log here. Error is eventually elevated to service status description, where a user could remediate the version error.

manager/manager.go
```diff
@@ -885,21 +889,80 @@ func (m *manager) getReqServices(service model.Service, plugin model.Plugin) (ou
...
+				return false, err
 			}
-			if serviceMap[k] < int(req.Min) { //req.Min is an int32, thanks Swagger
-				return false
+		}
+		if _, ok := serviceMap[nameConstraint]; !ok { //no services using this plugin req
+			log.Error("No plugin by name constraint in sigma")
```
adamschaub: Possibly spammy log. Esp on error level.

manager/manager.go
adamschaub: Nevermind, we removed all info about the plugin from the service. I'm not a huge fan, but it makes this point moot.

manager/manager.go
```diff
@@ -885,21 +889,80 @@ func (m *manager) getReqServices(service model.Service, plugin model.Plugin) (ou
...
+				checkingVer, err := version.NewVersion(p.Version)
+				if err != nil {
+					log.Error(err)
+					return false, err
+				}
+				if versionConstraint.Check(checkingVer) {
+					totalReached = totalReached + 1
+				}
+			} else { //if req has no version constraint, count all services for this plugin name towards the TotalReached
+				totalReached = totalReached + 1
```
adamschaub: totalReached++?

manager/manager.go
```diff
@@ -885,21 +889,80 @@ func (m *manager) getReqServices(service model.Service, plugin model.Plugin) (ou
...
+			return false, nil
+		}
+		totalReached := 0
+		for _, svc := range serviceMap[nameConstraint] {
+			p, err := m.db.GetPlugin(svc.PluginRef.GetID().String())
+			if err != nil {
+				log.Error(err)
+				return false, err
 			}
+			if versionConstraint != nil {
```
adamschaub: This section is a little messy, but works. Would be nice to get things more left-aligned.

PR #405, opened 2017-11-01
Makes sure that services are not added that reference themselves

### LINKED ISSUE #403: Service selector must check self
Or purgatory will ensue

rfliegerallison:
### PR checklist

### Change log

Adds extra validation to make sure that added services don't reference themselves as requirements.

### Notes

Resolves #403


api/v1/server/validate.go
```diff
@@ -202,6 +206,20 @@ func IsValidSelector(sel string) bool {
 	return true
 }
 
+// ServiceReferencesSelf returns true iff the specified requirement string
+// references the service itself.
+func ServiceReferencesSelf(svcName string, req string) bool {
+	pairs := strings.Split(req, "&")
+	for _, p := range pairs {
+		// Note that whitespace and empty values are allowed
```
meganokeefe: only for labels i think -- i don't think we're allowing whitespace and empty values for a service name?

PR #410, opened 2017-11-02
Plugin logger for standardized logging

### LINKED ISSUE #355: Plugin log standardization doc + fix examples
We want logs to all look the same across plugins.

Should update developer guide too once it's done.

rfliegerallison:
### PR checklist

### Change log

Added a plugin `logger` in `sigma/plugin/logger` which needs to be initialized with `self` at the beginning of each plugin implementation.

Also updates examples to use this new standard.

### Notes

Resolves #355 

### Todo

Update `latest` containers to have these changes.


meganokeefe:
Looking at some of the vendor changes, what in `coffee-orders` uses `vendor/golang.org/x/` ?

log/fluentd/Dockerfile
```diff
@@ -1,3 +1,4 @@
 # fluentd/Dockerfile
 FROM fluent/fluentd:v0.12-debian
 RUN ["gem", "install", "fluent-plugin-elasticsearch", "--no-rdoc", "--no-ri", "--version", "1.9.2"]
+RUN ["gem", "install", "fluent-plugin-rename-key", "--no-rdoc", "--no-ri", "--version", "0.3.4"]
```
meganokeefe: can you add something along the lines of  `docker rmi log-fluentd:latest` to `log/deploy.sh`?

log/fluentd/conf/fluent.conf
```diff
@@ -1,9 +1,26 @@
...
   bind 0.0.0.0
 </source>
+
+<filter *.**>
+  @type parser
+  format json
+  key_name log
+  reserve_data true
+  suppress_parse_error_log true
+  ignore_key_not_exist true
```
meganokeefe: Re: this filter-- so if someone is writing a golang plugin but not using the plugin logger, nad they don't have a field `log` for instance, what will happen to that log message? 

Applies to non golang plugins as well.

log/fluentd/conf/fluent.conf
```diff
@@ -1,9 +1,26 @@
...
+</filter>
+
+<filter *.**>
+  @type rename_key
+  rename_rule1 ^msg$ message
+  rename_rule2 ^log$ raw
```
meganokeefe: Nice nice

plugin/logger/logger.go
```diff
@@ -0,0 +1,309 @@
...
+)
+
+// PluginLogger contains all information needed for plugin logging.
+type PluginLogger struct {
+	logr *logrus.Logger
+	self logrus.Fields
+}
+
+// New initializes a new PluginLogger.
+func New(self *client.Self) *PluginLogger {
```
meganokeefe: Reflecting on the Text Vs. Json dilemma, in theory ..... we could change the `self()` API call to return the sigma log config. That would be pretty janky but it would allow us to set the `logr.Formatter` to the right thing and not have json-formatted Stdout all the time. 

https://github.com/sigma-dev/sigma/issues/420

meganokeefe: NVM we are #layingthistorest

log/fluentd/Dockerfile
rfliegerallison: So you want this to always happen whenever we deploy? The container with always need to get rebuilt?

log/fluentd/conf/fluent.conf
rfliegerallison: If there's no `log` field, the entry should get skipped by the filter. If there is a log field, and the format is JSON, it will parse out the fields and then rename the log field to `raw`.

log/fluentd/Dockerfile
mikeynap: wouldn't docker pull log-fluentd:latest accomplish the same thing?

mikeynap:
I'm ok with leaving the vendor.json file in these examples, but having the vendor folders is ridiculous.

PR #423, opened 2017-11-14
Event Driven Manager

stephenrlouie:
### Change log

Refactor of the manager to be event driven instead of constantly cycling through existing services. This means API calls will add events into the RabbitMQ queue and the manage is simply reading `Deploy` `Destroy` `Update` `Status` and `Finish` off that queue to process all events and track state.

I kept the state machine simple because I wanted to only change the events functionality here. We could add an `updating` state, but right now we have 3 states` Inactive` `Active` and `Failed`. I'm not convinced a more robust state machine gets us a lot more. We can already know if services are not in a desired state, by checking the current state and the fields `Remove` and `Disable`

## Event Types
`Destroy`: Takes a services and stops it, possibly removes from the DB
`Deploy`: Takes an inactive or failed services and attempts to start it
`Status`: Will attempt to status on `Unknown` or `Active` services, this happens on a ticker for each service.
`Update`: Can trigger 3 types of actions, `Update`,`Destroy` (disable, remove), or `Deploy` (enable).
`Finish`: Means we've received a response from the plugin and will process it and close out the ServiceOperation
`Queue`: Means the manager will look through all services and determine if a Status is required. It will also be a fall back mechanism to ensure services `desired.State == current.State`. It will also update the `Metadata.Requirements` checking if a new `req` exists and if one is found we will queue an `Update` for that service.

## Validated / Done 
- Single services that do not have dependencies and properly call back.
- Simple hello -> goodbye dependency. with updates and having goodbye exist before hello.
- Cascade update, 1 level only right now.
- Fixed: channel-init breaks, spams events (manager was wrong, not channel-init)

#### existing issues 
- Handle a service that does not call back, we must timeout on a service after some time. (this should be triggered by a status) (*megan is doing*) 
- tracking tasks and failing ones that do not call back. [DEFINITELY NOT WORKING.](https://github.com/sigma-dev/sigma/blob/events/manager/manager.go#L528) 
- double post to ServiceOpOutput causes problems.
- Double status in the queue after a destroy, I don't think I'll clear the status from the queue.
- Multiple events of same type in queue, always trying to startOp. How do we know what we can throw out?  (DB is updated. Or ordering, if we have 10 updates back to back, does order matter?) As long as DB is in final state, we just need to run one update.
- ensure DB is updated first in API calls, need to ensure all events can be triggered from status looper. Not just start and stop, need to implement an update catch.
- remove site and service explicit garbage collection, only let event handlers remove services
- better backoff for retry - nacks happen instantly + because rabbit puts nacks on the front of the queue. sleep before nack happens? 

### STEVE TODO 
- [x] fix logic to handle very short running services (eg. goodbye-world)
- [ ] make plugin client more strict **(issue created)**
- [ ] events have uuids; service operations get uuids  **(issue created)**
- [ ] self() gets my event UUID **(issue created)**
- [x] modify state machine for service operations 
- [x] be better about NACKing (sleep?) 

### general TODO 
- [ ] if a service is destroyed successfully, but there are other events (eg. status) for that service on the queue, we should have logic to handle that **(issue created)**


api/v1/server/server.go
```diff
@@ -596,15 +606,23 @@ func (s *Server) AddServiceOperationOutput(w http.ResponseWriter, r *http.Reques
...
+			return http.StatusBadRequest, errors.New("Invalid Status")
+		}
+
+		op2Update.Status = apiObj.Status
 		op2Update.Output.Status = apiObj.Status
 		op2Update.Output.Vars, _ = apiObj.Vars.(map[string]interface{})
 
-		// TODO: make sure this status is an accepted enum
-		op2Update.Status = apiObj.Status
+		err = s.eventProducer.Send(queue.Event{SvcID: op2Update.ServiceRef.GetID().String(), Operation: model.ServiceOperationFinish})
```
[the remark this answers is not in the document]
adamschaub: Same goes for other operations. We probably want to latch in any updates before launching the event.

manager/service_provider
```diff
@@ -0,0 +1,107 @@
+// //Package manager controls the services and is how sigma interacts with services.
```
adamschaub: delete

queue/channels.go
```diff
@@ -1,30 +1,24 @@
 package queue
 
-import "time"
+import (
+	"time"
+)
```
adamschaub: no

manager/manager.go
```diff
@@ -47,7 +49,7 @@ var (
 	ErrServiceNotFound           = errors.New("Service not found")
 	ErrFromCycleDetected         = errors.New("Cyclical froms detected.")
 	ErrOperationHeartbeatExpired = errors.New("Operation heartbeat expired")
-	lockTime                     = 10 * time.Second
+	LockTime                     = 1 * time.Minute
```
adamschaub: why export?

manager/manager.go
```diff
@@ -151,35 +147,47 @@ func NewManager(db database.Database,
...
+	go m.statusTicker(m.serviceStatusInterval)
 	for {
 		select {
 		case <-m.quit:
 			log.Info("Waiting for workers to finish")
 			m.workerWg.Wait()
 			log.Info("Quitting Sigma Manager")
 			return nil
-		case <-gcTicker:
+		case <-gcTicker.C:
```
adamschaub: just note that with this ticker, you are trying to run every `m.gcInterval`, including the time it takes to garbage collect, handle events, etc. Previous impl was meant to keep the time between ticks constant. Just so you are cognizant of the change you're making.

manager/manager.go
```diff
@@ -189,182 +197,566 @@ func (m *manager) Start() error {
 					m.workerWg.Add(1)
 					go func() {
 						defer m.workerWg.Done()
-						m.updateService(svc)
-						err := m.svcConsume.Ack(svc)
-						if err != nil {
-							log.WithError(err).Errorf("Unable to ack service %v", svc)
+						var err error
```
adamschaub: this is all gross. Can we not encapsulate this in a func like `updateService` so that the logic for starting/stopping the worker isn't obfuscated.

manager/manager.go
```diff
@@ -189,182 +197,566 @@ func (m *manager) Start() error {
...
+			return nil
+		}
+		return err
+	}
+	var internalErr error
+	switch svc.Status.State {
+	case model.ServiceStatusFailed:
+		var op string
+		if svc.IsDisabled() || model.BoolValue(svc.Remove) {
+			op = model.ServiceOperationDestroy
```
adamschaub: Why run a destroy if it has failed? Wasn't our previous strategy to run deploy again, until we failed a max # of times?

api/v1/server/server.go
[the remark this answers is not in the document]
stephenrlouie: I think we need events first, because if we update a DB and don't kick off the event no one will ever get around to handling that event. 

If an event is in the queue by mistake we should just pass on it and move on. When we get back to retry the DB update it can make sure the event will get through the next time.

manager/service_provider
stephenrlouie: yea I left this as a note just to see how things went before

queue/channels.go
stephenrlouie: thats some golinter I have on, my b

manager/manager.go
stephenrlouie: hmm truth, I just thought this was simpler. didn't think that distinction was a huge deal, maybe it is.

manager/manager.go
stephenrlouie: umm yes, this is vomit.

manager/manager.go
stephenrlouie: if an update to a service is made and it has `disabled` set to true or `remove` set to true. we should destroy.

database/deploy.sh
```diff
@@ -8,7 +8,7 @@ fi
 docker pull couchbase:4.6.3
 docker run -d --name sigma-db  -p 8091-8094:8091-8094 -p 11210:11210 $USEHOST couchbase:4.6.3
 COUNT=0
-LIMIT=15
+LIMIT=25
```
adamschaub: bad limit, make it BIGGER

stephenrlouie:
Please find TODOs in code and the checkboxes in this PR message. Its not perfect, but I think it works for good plugins and workflows.

meganokeefe:
log-levels: might want to reduce some of these to `debug` at some point. 

```
INFO[2017-12-05T18:46:24-05:00] Deploy Event Received                         service-id=ad346cf4-5566-4d7e-b2bb-066fe8245768
INFO[2017-12-05T18:46:27-05:00] Started Operation                             operation-id=ad346cf4-5566-4d7e-b2bb-066fe8245768-opID operation-type=Deploy service-id=ad346cf4-5566-4d7e-b2bb-066fe8245768
INFO[2017-12-05T18:46:29-05:00] Finish Event Received                         service-id=ad346cf4-5566-4d7e-b2bb-066fe8245768
INFO[2017-12-05T18:46:29-05:00] Service Active.                               Service=helloSvc
INFO[2017-12-05T18:46:36-05:00] Queue Event Received                          file=manager.go line=224
```

manager/manager.go
```diff
@@ -25,19 +25,22 @@ import (
...
-	DefaultMinOperationStatusInterval = "100ms"
-	DefaultStatusInterval             = "100ms"
-	DefaultUpdateInterval             = "100ms"
-	DefaultRetryInterval              = "5s"
-	DefaultMaxWorkers                 = 10
-	DefaultMaxRetries                 = 10
-	DefaultGCInterval                 = "1m"
-	archiveLimit                      = 10
+	DefaultCheckQueueInterval = "10s"
+	DefaultStatusInterval     = "5s"
```
meganokeefe: I get that these defaults were maybe increased for debugging purposes but we should probably reduce them again - like `defaultstatusinterval` - before merge

manager/manager.go
```diff
@@ -47,7 +50,8 @@ var (
 	ErrServiceNotFound           = errors.New("Service not found")
 	ErrFromCycleDetected         = errors.New("Cyclical froms detected.")
 	ErrOperationHeartbeatExpired = errors.New("Operation heartbeat expired")
-	lockTime                     = 10 * time.Second
+	ErrReqFetch                  = errors.New("Failed to get requirements")
+	LockTime                     = 1 * time.Minute
```
meganokeefe: Why was `LockTime` increased x6?

manager/manager.go
```diff
@@ -151,220 +149,812 @@ func NewManager(db database.Database,
...
+		m.workerLock.Lock()
+		if m.workerCount < m.maxWorkers {
+			m.workerCount++
+			m.workerWg.Add(1)
+			go func() {
+				defer m.workerWg.Done()
+				var err error
+				switch ev.Operation {
+				case model.ServiceOperationDeploy:
+					log.WithField(model.LogSvcID, ev.SvcID).Infof("Deploy Event Received")
```
meganokeefe: debug these logs

manager/manager.go
```diff
@@ -151,220 +149,812 @@ func NewManager(db database.Database,
...
+		if err == ErrMinReqsNotMet {
+			return nil
+		}
+		return err
+	}
+
+	var internalErr error
+	switch svc.Status.State {
+	case model.ServiceStatusFailed:
+		err, err2 := m.retryOperation(&svc, plg, vars, model.ServiceOperationDeploy)
```
meganokeefe: is there a way to clean up this `err` / `err2` logic and possibly combine the two errs? or rename them if they're logically different?

manager/manager.go
```diff
@@ -151,220 +149,812 @@ func NewManager(db database.Database,
...
+				internalErr = err
+			}
+		}
+	default:
+		log.Warnf("Deploy can only be used on Services with Status: %q/%q. Received: %q",
+			model.ServiceStatusInactive, model.ServiceStatusFailed, svc.Status.State)
+		return nil
+	}
+
+	idList := make([]string, len(reqs))
```
meganokeefe: add a comment here for what lines 313-320 do

manager/manager.go
```diff
@@ -151,220 +149,812 @@ func NewManager(db database.Database,
...
 	if err != nil {
 		return err
 	}
-	for _, id := range siteIDList {
-		myself, err := m.db.GetSite(id)
+	err = m.finishOperation(op)
+	if err != nil {
+		return err
+	}
+	err = m.archiveOperation(op)
```
meganokeefe: nit you can just return on this line

manager/manager.go
```diff
@@ -151,220 +149,812 @@ func NewManager(db database.Database,
...
+	}
+
+	return nil
+}
+
+// Status event
+func (m *manager) statusEventHandler(svcID string) error {
+	svc, plg, _, vars, err := m.getServiceDetails(svcID)
+	// don't want to block a status from happening if reqs are messed up
+	if err != nil && err != ErrMinReqsNotMet {
```
meganokeefe: edge case: what if deploy happens for a service, service is active, triggering a status. but then you destroy one of the service's reqs, such that ErrMinReqsNotMet == true now. Do we status in this case or put the service back into `Inactive`?

manager/manager.go
```diff
@@ -151,220 +149,812 @@ func NewManager(db database.Database,
...
+		return internalErr
+	}
+
+	return nil
+}
+
+// TODO: get all running service operations and check their status.
+// Retire one's that have not called back
+// TODO: More robust to catch state inconsistency
+func (m *manager) queueEventHandler() error {
```
meganokeefe: Can you break this function into helpers? it's > 100 lines haha

manager/manager.go
```diff
@@ -151,220 +149,812 @@ func NewManager(db database.Database,
...
+			} else {
+				log.WithField(model.LogSvcID, s.GetID().String()).Info("Vars changed! Updating service")
+				return nil
+			}
+		}
+
+		// checks for new reqs
+		err = m.updateReqs(s.GetID().String())
+		if err != nil && err == ErrMinReqsNotMet {
+			return nil
```
meganokeefe: but what if reqs suddenly are not met? should we make this service  status=inactive here?

manager/manager.go
```diff
@@ -151,220 +149,812 @@ func NewManager(db database.Database,
...
+		return err
+	}
+
+	// Updates reqs, might queue and update
+	// Looks are vars and previous vars, might queue an update
+	// looks at state might destroy or update
+	// I'd guess we should only queue one of these events per time around. To try and keep the
+	// queue pressure to a min
+	for _, s := range svcs {
+		// See active services
```
meganokeefe: put everything inside this `for` into a helper plz

also, is it worth running checks on all services in parallel AKA spawning this helper in a goroutine? 
what is there are a ton of services you're looping on?

manager/manager.go
```diff
@@ -151,220 +149,812 @@ func NewManager(db database.Database,
...
+		log.WithError(err).Errorf("Failed to get req")
+		return err
+	}
+
+	idList := make([]string, len(reqs))
+	for i, r := range reqs {
+		idList[i] = r.GetID().String()
+	}
+
+	// TODO functionalize this stuff
```
meganokeefe: yes please

manager/manager.go
```diff
@@ -151,220 +149,812 @@ func NewManager(db database.Database,
...
+			if err != nil {
+				log.WithError(errors.New("Failed to update new reqs on update queued")).Errorf("Update SVC problem")
+			}
+		}
+	}
+
+	return nil
+}
+
+func (m *manager) checkOperation(op model.ServiceOperation) {
```
meganokeefe: comment func header please

cmd/sigma.go
```diff
@@ -74,18 +74,18 @@ func sigmaRun(cmd *cobra.Command, args []string) {
 	}
 
 	// Queues
-	var svcProvide queue.ServiceProvider
-	var svcConsume queue.ServiceConsumer
+	var eventProducer queue.EventProducer
+	var eventConsumer queue.EventConsumer
 	if gCfg.UseChannelInit {
 		channelQueue := queue.NewChannelQueue(100)
```
meganokeefe: can we make `100` a default here / const?

api/v1/server/server.go
```diff
@@ -1472,6 +1498,7 @@ func (s *Server) UpdateService(w http.ResponseWriter, r *http.Request) {
 			return http.StatusForbidden, ErrServiceRemoved
 		}
 
+		updatedService.Status = service2Update.Status
```
meganokeefe: No!!!!!!!!

Don't do this!!!!!!

Bad

no

manager/manager.go
```diff
@@ -759,13 +1371,26 @@ func (m *manager) collectOutput(svc *model.Service, output map[string]interface{
...
-			return ErrOperationExists
+			// Have multiple operaitons of the same type coming in and starting.
+			// Not sure a more complex state machine fixes this
+			if opType == dbOp.Operation {
+				//Start op because its the same operation type as me
+				// TODO what if we have two updates, we will try to start another one no top of an existing?
+				// I think we have to return error, to double update. wait for the current one to finish
+				// If we have 10 updates in a row, how do we order them properly? if we have concurrent updates all going.
+				return m.startOperation(dbOp)
+			} else if opType != dbOp.Operation {
```
meganokeefe: this can just be an `else` yes?

manager/manager.go
stephenrlouie: I didn't want to release the lock to soon. All of queue needs to happen in this time.

manager/manager.go
stephenrlouie: I believe #469 fixed. should double check

PR #424, opened 2017-11-14
Fixes edge case causing PluginLogger to panic

### LINKED ISSUE #421: Plugin logger panic
##### Description

If anything causes a new PluginClient to not be initialized properly, we can't use the plugin logger because it hasn't been initialized yet. We need to simply print to STDERR.

rfliegerallison:
### PR checklist

### Change log

Fixes logging in the main function so the logger isn't called when it failed to be instantiated.

### Notes

Resolves #421 

### TODO

`latest` containers need to be updated to use these changes.


examples/coffee-run/coffee-orders/main.go
```diff
@@ -119,10 +121,10 @@ func (c *CoffeeOrders) update(op pc.PluginArg) error {
 
 func main() {
 	c, err := New()
-	c.log.Println("Welcome to the Coffee Orders Plugin")
 	if err != nil {
-		c.log.Fatalf("Failed to Initialize: %v", err)
+		fmt.Fprintf(os.Stderr, "Failed to Initialize: %v", err)
```
mikeynap: gotta add an exit here or else we'll still panic :0

examples/coffee-run/coffee-prices/main.go
```diff
@@ -136,10 +138,10 @@ func (c *CoffeePrices) update(op pc.PluginArg) error {
 
 func main() {
 	c, err := New()
-	c.log.Println("Welcome to the Coffee Prices Plugin")
 	if err != nil {
-		c.log.Fatalf("Failed to Initialize: %v", err)
+		fmt.Fprintf(os.Stderr, "Failed to Initialize: %v", err)
```
mikeynap: Unfortunately I think you gotta do this for all the plugins... feels bad man

examples/coffee-run/coffee-orders/main.go
rfliegerallison: if i `return` that should be fine, right?

PR #429, opened 2017-11-14
Log Query Endpoint

mikeynap:


log/fluentd/Dockerfile
```diff
@@ -2,3 +2,4 @@
 FROM fluent/fluentd:v0.12-debian
 RUN ["gem", "install", "fluent-plugin-elasticsearch", "--no-rdoc", "--no-ri", "--version", "1.9.2"]
 RUN ["gem", "install", "fluent-plugin-rename-key", "--no-rdoc", "--no-ri", "--version", "0.3.4"]
+RUN ["gem", "install", "fluent-plugin-concat","--no-rdoc", "--no-ri", "--version", "1.0.0"]
```
rfliegerallison: Kill this

log/fluentd/conf/fluent.conf
```diff
@@ -21,6 +21,13 @@
   rename_rule2 ^log$ raw
 </filter>
 
+<filter *.**>
```
rfliegerallison: Kill

log/log.go
```diff
@@ -90,6 +92,9 @@ func NewLogr(conf Config) error {
 		if makeErr != nil {
 			fmt.Println(makeErr)
 		}
+		if conf.ElasticSearchURL != "" {
+			Query, _ = NewQuerier(conf.ElasticSearchURL)
```
rfliegerallison: Is there a good reason why ignore the returned `error`? I'd check it

log/query.go
```diff
@@ -0,0 +1,66 @@
+package log
+
+import (
+	"errors"
+	"reflect"
+	"strings"
+
```
rfliegerallison: NIT; delete this line and keep the built-in Go packages grouped together

log/query.go
```diff
@@ -0,0 +1,66 @@
...
+		bq = bq.Must(elastic.NewMatchQuery("service_id", serviceID))
+	}
+	if len(opID) > 0 {
+		bq = bq.Must(elastic.NewMatchQuery("operation_id", opID))
+	}
+
+	searchResult, err := q.Client.Search().
+		//FetchSource(true).
+		Query(bq).             // specify the query
+		From(from).Size(size). // take documents 0-9
```
rfliegerallison: I assume this comment refers to the default batch size being 10? maybe update the comment to `take documents 0-(size-1)` or `take documents [0, size)`? something general like

api/v1/server/server_test.go
```diff
@@ -208,7 +208,7 @@ func TestServiceOperationLogs(t *testing.T) {
 		Logs: "hello world log",
 	}
 	client.AddServiceOperationLogs(svcName, useID, o)
-	got, _, _ := client.GetServiceOperationLogs(svcName, useID)
+	got, _, _ := client.GetServiceOperationLogs(svcName, useID, 0, 100)
```
rfliegerallison: I would catch the error returned here and do something with it, just to be safe

PR #432, opened 2017-11-16
Slice query fix for Couchbase 

### LINKED ISSUE #430: Couchbase Query Services.Requirements
##### Description

Events need the ability to find all the services that are using me as a dependency. This requires me to query the DB based on if `Service.Requirements` contains my identifier

##### Sigma Version
0.3.2

##### Sigma Configuration
One using couchbase

##### Steps to reproduce

Create a service that has requirements, attempt to query on them. Via API QueryServices endpoint.

##### Suggest A Fix

A query could look like. `SELECT * from default WHERE Metadata.Type="SERVICE" AND ARRAY_CONTAINS(Requirements, "service=helloSvc")` but it must be able to handle the variety of labels that can exist in the requirements list.

[QueryServices](https://github.com/sigma-dev/sigma/blob/master/database/couchbase.go#L401) needs an update to be able to handle arrayContains

meganokeefe:
Closes #430  

### problem 

querying requirements worked for objects but not for filters. We weren't parsing a URL with filters like `http://127.0.0.1:24601/v1/services?requirements=[service=helloSvc]` correctly for couchbase. I added these logs in `makeQuery` -- 

`INFO[2017-11-16T09:14:24-05:00] FILTER QUERY - fieldName=Requirements, kind=string` 

vs. 

`INFO[2017-11-16T09:14:24-05:00] OBJ FIELD QUERY - fieldName=Requirements, kind=slice` 

### solution 

I made the fix on the API side and didn't end up having to change couchbase at all. I unmarshal from string the same way we deal with attribute filters in a URL. 

Also note that each string in the array has to be wrapped (or escaped) with quotes -- in postman looks like this: 

screen shot 2017-11-16 at 9 39 27 am


stephenrlouie:
tests please

PR #441, opened 2017-11-21
Adds ability to group consecutive logs into a single batch (useful for Ansible output)

rfliegerallison:
### PR checklist

### Change log

Updates fluent.conf and plugin logger to support log concatenation.

### TODO

Get it to work.


log/fluentd/Dockerfile
```diff
@@ -2,3 +2,4 @@
 FROM fluent/fluentd:v0.12-debian
```
meganokeefe: Slightly unrelated, but could you add a line in `log/deploy.sh` to do a `docker rmi log_fluentd`? This way any future updates to this Dockerfile won't be clobbered by an existing `log_fluentd` image. Thanks!

plugin/logger/logger.go
```diff
@@ -9,6 +9,16 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+var (
```
meganokeefe: I would make these `const` instead of `var`, since the fluentd conf relies on a regex with exactly this

log/fluentd/Dockerfile
rfliegerallison: wouldn't that mean docker would have to rebuild a fresh image every time logs are deployed? i feel like we don't want that..

plugin/logger/logger.go
rfliegerallison: good call

log/fluentd/Dockerfile
meganokeefe: Yeah i guess you're right - you can hold off for now then.

PR #446, opened 2017-11-27
Selectors are Alphabetized on AddService and GetServices

### LINKED ISSUE #439: Query by Requirements takes string literal
Querying by requirements may be too literal to be useful. Each selector line in `Requirements` is a string. We look for an exact string match when we query the DB based on `Requirements` We need to either be consistent on order of selectors and wrap that so users can Add / Query in any order and still get the expected results.

`plugin=testing:1.1&service=byebyebye` will only match that string. It will not match another service initialized with `service=byebyebye&plugin=testing:1.1&`

meganokeefe:
Closes #439 

This allows the user to list their selectors in any order. For example, if you have this service with requirements: 

```
	"requirements": [
		"plugin=hello:0.3&site=central"
	],
```

You can query with `http://10.131.32.69:24601/v1/services?requirements=["site=central%26plugin=hello:0.3"]` or `http://10.131.32.69:24601/v1/services?requirements=["plugin=hello:0.3%26site=central"]` and still get the service back. 

The internals work as follows:
- We alphabetize each requirements' selectors on service add, in `convert.go`. So in the database, reqs are always alphabetized even though the user doesn't know it.
- On query with filters or with an object, we alphabetize the selectors again so that the requirements query is deterministic -- allowing us to do a strict string match w/ what's in the DB.


config.yaml
```diff
@@ -11,7 +11,7 @@ Manager:
 #  ClientUrl: 127.0.0.1:4646
 #  FluentdAddr: 127.0.0.1:24224
 Docker:
-  APIUrl: 127.0.0.1:24601
```
stephenrlouie: please remove your IPs

PR #448, opened 2017-11-28
Improved validation for Service Requirements

meganokeefe:
Checks plugin name:version format, and better errors for `isValidSelector()`.


stephenrlouie:
tests please

PR #459, opened 2017-12-11
Events - Bug Fix PR

meganokeefe:
### pre review changelog 
- [x] Cleaned up the `queueEventHandler` logic to be more easily understandable
- [x] Fixed a bug in the operation state machine that caused the heartbeat check to never happen
- [x] Have verified that plugins that never call home (eg. sleep for an hour past the heartbeatTTL) are gracefully destroyed and redeployed by manager
- [x] Fixed a bug where updating a dependency service breaks all the services in the chain. Basically if an update comes in from the queue and another op of a different type (eg. StatusOperation) is running, we treat the service as failed. Vs. trying again.
- [x] Added a new example plugin, `sleeper`, that is a real plugin, produces `hello world` style logs, etc. but has the option of 1) never heartbeating, 2) never sending outputs, 3) both. (via variables. AKA you can trigger a failure if you want).
- [x] Added another example plugin, `hello-again`, that relies on goodbye. Allows us to test a chain of 3 small plugins --> update propagation. it doesn't have to stay, but it's good for debugging.

### NEW changes 
- [x] in `finishoperation`, `queueChildren` shouldn't happen on a destroy
- [x] `finishUpdate` code is populated when a service is in status `Failed`
- [x] ^^ relatedly `finishUpdate` should allow a transition into `servicestatusinactive`, just like `finishDeploy`.
- [x] made the new `invalidStateTransition` warn for very short running services a debug, not a warn
- [x] fixed a logic bug where, if you delete a service's dependency, it still runs status checks and stays Active. now i'm disabling it to inactive


manager/manager.go
```diff
@@ -379,14 +380,19 @@ func (m *manager) updateEventHandler(svcID string) error {
 				internalErr = err
 			}
 		} else if m.isReadyForUpdate(reqs) {
+			existingOps, _ := m.db.QueryServiceOperations(model.ServiceOperation{ServiceRef: model.ModelRef(svc.GetID())}, nil)
```
stephenrlouie: we probably want this for every event type, so we can move this to dispatch workers. and we will want to NACK events, so it does't get removed from the queue

manager/manager.go
```diff
@@ -523,87 +529,117 @@ func (m *manager) queueEventHandler() error {
 		return err
 	}
 
-	// Updates reqs, might queue and update
-	// Looks are vars and previous vars, might queue an update
-	// looks at state might destroy or update
-	// I'd guess we should only queue one of these events per time around. To try and keep the
-	// queue pressure to a min
+	// Take action on all services.
+	// TODO - would like to make this a goroutine but does statusLock prevent this?
```
stephenrlouie: we can make this concurrent. As long as all of them finish their work and we release the lock at that point. Its an optimization we don't need this second

manager/manager.go
```diff
@@ -523,87 +529,117 @@ func (m *manager) queueEventHandler() error {
...
+func (m *manager) processService(s model.Service) error {
+	var curOpStatus string
+
+	// Get op for this service
+	op, getOpErr := m.db.GetServiceOperation(model.OperationID(s.GetID().String()))
+	if getOpErr != nil && getOpErr != database.ErrNotFound {
+		return getOpErr
+	}
+
+	if getOpErr != database.ErrNotFound {
```
stephenrlouie: how does this ever get run, wouldn't it get pulled into the above if first

manager/manager.go
```diff
@@ -523,87 +529,117 @@ func (m *manager) queueEventHandler() error {
...
+	}
 
-		// see's if we have to status
-		if t, _ := s.GetTime(model.SigmaMetaTimeLastStatus); (t == (time.Time{}) || t.Add(m.serviceStatusInterval).Before(time.Now())) && s.Status.State == model.ServiceStatusActive {
-			err := m.eventProducer.Send(queue.Event{SvcID: s.GetID().String(), Operation: model.ServiceOperationStatus})
+	// Determine if we have to force-kill or redeploy any failed service operations for this svc
+	switch s.Status.State {
+	case model.ServiceStatusInactive:
+		// 2 cases where we'd want to re-deploy -- in failing state, or inactive AND no service op
+		if (getOpErr != nil && getOpErr == database.ErrNotFound) || curOpStatus == model.ServiceOperationStatusFailed && (s.IsDisabled() == model.BoolValue(s.Remove)) && !s.IsDisabled() {
```
stephenrlouie: do we need this getOpErr again?

manager/manager.go
```diff
@@ -523,87 +529,117 @@ func (m *manager) queueEventHandler() error {
...
-				err := m.eventProducer.Send(queue.Event{SvcID: s.GetID().String(), Operation: model.ServiceOperationUpdate})
-				if err != nil {
-					log.WithField(model.LogSvcID, s.GetID().String()).WithError(err).Error("Failed to put update on queue")
-					return err
-				}
-				log.WithField(model.LogSvcID, s.GetID().String()).WithError(err).Warn("Service failed - update operation queued")
-				return nil
+	case model.ServiceStatusFailed:
+		stat := s.Status.State
+		// This might be tricky, bc we might have a disable event in the queue that hasn't landed. I guess DB needs to be up'd
```
stephenrlouie: disable event?

manager/manager.go
```diff
@@ -523,87 +529,117 @@ func (m *manager) queueEventHandler() error {
...
+		// TODO Need this to go through, if it does not we will not update, unless we have prev-reqs and current.
+		// We will loop around to this when the queue event comes around again, maybe not a big deal
+		if err != nil {
+			log.Errorf("Failed to queue Update operation: %v", err)
+		} else {
+			log.WithField(model.LogSvcID, s.GetID().String()).Debug("Vars changed! Updating service")
+			return nil
+		}
+	}
+	err = m.updateReqs(s.GetID().String())
```
stephenrlouie: I think we want to update the DB first? I had this debate with Adam and Matt back when I was working on this originally. We just need to make sure if we add to DB, we will still trigger the event into the queue.

manager/manager.go
```diff
@@ -658,14 +694,16 @@ func (m *manager) updateReqs(svcID string) error {
...
+	if needsUpdate {
+		log.Warnf("Service=%s needs an update! Queueing one", svc.GetName())
 		// Set previous reqs
 		svc.Set("prev-reqs", currStrings)
 		svc.Set("reqs", idList)
 		err = m.db.UpdateService(svc)
 		if err != nil {
 			log.WithError(err).Error("Fails to up reqs")
 		}
+		return m.eventProducer.Send(queue.Event{SvcID: svc.GetID().String(), Operation: model.ServiceOperationUpdate})
```
stephenrlouie: not using the QueueUpdate function? (Should we wrap the adding events to the queue?)

manager/manager.go
meganokeefe: ^ addressing that first comment here 

^ you are addressing the second comment yes?

manager/manager.go
meganokeefe: Ok, putting an issue out

examples/triplethreat.sh
```diff
@@ -0,0 +1,13 @@
+SIGMA_HOST=${SIGMA_HOST:=localhost}
```
stephenrlouie: remove?

config.yaml
```diff
@@ -4,14 +4,13 @@ Log:
...
-  UpdateInterval: 5s
-  OperationUpdateInterval: 5s
-  ServiceStatusInterval: 20s
+  ServiceStatusInterval: 10s
 #Nomad:
 #  ClientUrl: 127.0.0.1:4646
 #  FluentdAddr: 127.0.0.1:24224
 Docker:
-  APIUrl: 127.0.0.1:24601
+  APIUrl: 10.131.35.234:24601
```
stephenrlouie: meegan

manager/manager.go
```diff
@@ -523,87 +536,115 @@ func (m *manager) queueEventHandler() error {
...
+	return nil
+}
+
+// Takes action on a service if needed (desired state -> actual state)
+func (m *manager) processService(s model.Service) error {
+	var curOpStatus string
+
+	// Get op for this service
+	op, getOpErr := m.db.GetServiceOperation(model.OperationID(s.GetID().String()))
+	if getOpErr != nil && getOpErr != database.ErrNotFound {
```
stephenrlouie:
```
if  getOpErr == nil {
    curOpStatus = m.checkOperation(op)
} else if getOpErr != database.ErrNotFound {
   return getOpErr
}

File: manager/manager.go [hunk 11]
```diff
@@ -523,87 +536,115 @@ func (m *manager) queueEventHandler() error {
...
-			err := m.eventProducer.Send(queue.Event{SvcID: s.GetID().String(), Operation: model.ServiceOperationStatus})
+	if getOpErr == nil {
+		curOpStatus = m.checkOperation(op)
+	}
+
+	// Determine if we have to force-kill or redeploy any failed service operations for this svc
+	switch s.Status.State {
+	case model.ServiceStatusInactive:
+		// 2 cases where we'd want to re-deploy -- in failing state, or inactive AND no service op
+		if (getOpErr != nil && getOpErr == database.ErrNotFound) || curOpStatus == model.ServiceOperationStatusFailed && (s.IsDisabled() == model.BoolValue(s.Remove)) && !s.IsDisabled() {
```
stephenrlouie: just need the `getOpErr == database.ErrNotfound`

stephenrlouie: thinking about it, we shouldn't have to know if there is a status operation. if this comes around and we see its not active and wants to be. we just put the deploy event on the queue. Our event processor should be able to handle a multiple deploy events on the queue

manager/manager.go
stephenrlouie: nvm, I actually think this is right. Because we detect a change by seeing a difference, once we hit the DB with this updateReqs, we wouldn't be able to detect that an Update is required. We could do this with an updateFlag / updateState so we know an update is required.

manager/manager.go
```diff
@@ -1187,16 +1249,18 @@ func (m *manager) finishDeploy(op model.ServiceOperation, svc *model.Service) er
...
 	switch op.Status {
 	case model.ServiceOperationStatusFailed:
-
+		// TODO should something go here??
 	case model.ServiceOperationStatusSucceeded:
 		svc.Status.State = model.ServiceStatusActive
 		svc.Status.Description = ""
 		svc.PreviousVars = op.Vars
 	}
+	log.WithField("Service", svc.GetName()).Infof("Service Active.")
```
stephenrlouie: use the `model.Log` fields so we have standard log fields for ELK

manager/manager.go
```diff
@@ -695,53 +738,69 @@ func (m *manager) updateReqs(svcID string) error {
...
+func (m *manager) checkOperation(op model.ServiceOperation) string {
 	switch op.Status {
-	case model.ServiceOperationStatusCreated:
+	case model.ServiceOperationStatusStarted:
 		running, err := m.tasker.IsRunning(op)
 		if err != nil {
 			log.WithField(model.LogOpID, op.GetID().String()).WithError(err).Debugf("Unable to get status of operation")
 		}
 
 		if running {
```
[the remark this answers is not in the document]
meganokeefe: fixed

manager/manager.go
```diff
@@ -204,27 +205,40 @@ func (m *manager) DispatchWorkers() {
...
-					log.Debug("Queue Event Received")
-					err = m.queueEventHandler()
-				default:
-					log.WithField(model.LogSvcID, ev.SvcID).Warnf("Unexpected Event Received")
+
+				// For all handlers resposnible for starting service operations, must check that there isn't one running already
+				// TODO - don't just ack if an op already exists ? need event uuids to handle this in a more complex way?
+				handle := true
+				if ev.Operation != model.ServiceOperationQueue {
+					if m.serviceOperationExists(ev.Operation) {
```
stephenrlouie: this is the operation type.

meganokeefe: fixed this

